PR #1, opened 2022-10-13
WIP - discslab DLIO version

jovonho:
Opening this PR to show you the changes we have made to DLIO to use it to simulate our target workloads. We are still working on properly simulating BERT and DLRM. 

There are a lot of differences, in part from the use of the PyTorch branch that we integrated, and in part from features we added.

Starting from the PyTorch branch, we've added the following:
- Dockerfile with dependencies
- support for evaluation phases on a held-out dataset
- support for PyTorch DataLoader reading NPZ files
- microsecond precision logging
- scripts to launch DLIO in a container

These additions have only been tested for TFRecord files, and PyTorch Dataloader with NPZ files as this is what our workloads use. Because of this, other functionality is probably broken. More specifically, we're not trying to support the HDF5, CSV and NPZ readers, and we don't use Darshan or Tensorboard profiling.

I included the README which contains more information on these changes and current limitations. 
I think the most pressing point is the problem of PyTorch data_loader not being a data format in itself. We should probably reorganize this a bit as currently we have a npz Dataset included in data_loader_reader.py, but we'll need a new Dataset class for DLRM, so we might put them under a pt_datasets/ folder or something in that vein.

Feel free to contact me by email and we can also schedule calls to discuss.


README.md
```diff
@@ -1,3 +1,65 @@
...
+These additions have only been tested for TFRecord files, and PyTorch Dataloader with NPZ files as this is what our workload use. Because of this, other functionality is probably broken. More specifically, we're not trying to support the HDF5, CSV and NPZ readers, and we don't use Darshan or Tensorboard profiling.
+
+# Current Limitations
+
+There is an abtraction problem where DLIO considers PyTorch's data loader as a data type when it really is data type agnostic.
+Any of the existing data formats could be supported by extending the Dataset class and implementing the __len__ and __getitem__ methods. https://pytorch.org/docs/stable/data.html
+
+Since UNET3D reads in NPZ data, we currently only support this format for PyTorch. We will have to modify this for DLRM which reads in a custom binary format.
+
+Considering our objectives and the already significant changes made, I believe it might make more sense to change the abstraction to be workload based, instead of framework/datareader based.
```
hariharan-devarajan, line 22: This makes sense. I like the abstraction of having a framework. My recommendation is to have the following abstraction:
- Framework (switches between tf and pytorch)
- DataReader (framework agnostic data loading using iterators and yields)
- FrameworkDataLoader (framework specific loading for DataReader)

What do you think?

src/reader/data_loader_reader.py
```diff
@@ -0,0 +1,93 @@
...
+from src.common.enumerations import Shuffle
+from src.reader.reader_handler import FormatReader
+
+class DataLoaderReader(FormatReader):
+    """
+    DataLoader reader and iterator logic.
+    PyTorch data loader is file format agnostic so it is not technically separate from the other formats
+    For now we will assume we are reading .npz files through the dataloader
+    """
+    class NpzDataset(Dataset):
```
hariharan-devarajan, line 31: This class can move in DataReader. We can have a factory to load the correct data reader generically.

requirements.txt
```diff
@@ -1,6 +1,7 @@
-horovod[tensorflow]>=0.19.5
-tensorflow>=2.2.0
```
hariharan-devarajan, line 2: why was this removed?

src/reader/reader_handler.py
```diff
@@ -14,53 +17,101 @@ def __init__(self):
...
+            self._local_file_list_size = len(self._local_file_list)
+
             if read_shuffle:
                 random.shuffle(self._local_file_list)
+
+            logging.debug(f"{utcnow()} Rank {self.my_rank} will read {self._local_file_list_size} files: {self._local_file_list}")
+
+        # For PyTorch the file partitioning is done by the DistributedSampler in src/reader/data_loader_reader.py.
+        # We could also partition the files here like for TF and not use the sampler.
+        # We did it this way just to follow the UNET3D implementation as closely as possible.
```
hariharan-devarajan, line 106: I think this is interesting. We could decide how and when we want to partition based on a flag potentially to keep both versions and support more cases.

src/reader/tf_reader.py
```diff
@@ -73,24 +81,26 @@ def read(self, epoch_number):
...
-                break
+        dataset = self._dataset
+
+        # In tf, we can't get the length of the dataset easily so we calculate it
+        if self._debug:
+            total = math.ceil(self.num_samples*self._local_file_list_size/self.batch_size/self.comm_size)
+            logging.debug(f"{utcnow()} Rank {self.my_rank} should read {total} batches")
+
+        # The previous version crashed when all workers could not generate the same amount of batches
+        # Using the inbuilt tensorflow dataset iteration seems to work fine, was there an advantage of doing it the old way?
```
hariharan-devarajan, line 101: To have consistent iteration logic for each case. If we do TensorFlow iteration for one vs our own for others its would be inconsistent for format comparison.

hariharan-devarajan (commented):
Most changes are fine. I have some comments on abstractions and code.

requirements.txt
jovonho: It's now in the Dockerfile:
- tensorflow is obtained from the base image nvcr.io/nvidia/tensorflow:20.12-tf2-py3
- horovod is manually installed to have support for tf and pt (didn't manage to make it work in the requirements)

src/reader/data_loader_reader.py
jovonho: What would this look like, considering we already have npz_reader which is not the same as this? For all data formats we want to support with PT dataloader, we might need similar Dataset classes. See comment the README for more details.

README.md
jovonho: If I understand, we would have commands like `dlio --framework tf/pt --framework-dataloader y/n --format npz`. 

Depending on the `--framework-dataloader` flag, DLIO would use a TF Dataset, PT Dataloader, or a framework agnostic reader.

For file organization, we would keep the current `reader/` (minus `tf_reader.py` and `data_loader_reader.py`) and have a new folder like `framework_loader/` with a tf.data.Dataset-returning class and a torch.data.Dataloader-returning class, and ideally we would wrap the `reader/` classes inside.

I think this would be nice! However, I'm unsure of how easily we can wrap the `reader/` classes in the framework-specific data loaders. For example, the current `npz_reader` works great by itself, but to use it in a PyTorch Dataloader we would want to skip loading all the data in the `read()` method (to benefit from the on-demand loading aspect of data loaders), and thus couldn't simply wrap the `next()` method which yields the data assuming it has been read already. 

We could implement alternative paths etc, but the `npz_reader` class would then become complicated, when all that is needed to support npz in a PyTorch Dataloader is the 10-line NpzDataset we currently have in `data_loader_reader.py`.

Similarly, it might not be straight-forward to wrap `npz_reader` in a tf.data.Dataset.

So we may have to keep additional `pytorch_datasets/npz_dataset` and `tf_datasets/npz_dataset` classes, and do the same for each format we want to support.

jovonho:
Updated the PR with necessary work for v1.0 of the benchmark.

This includes UNET3D and BERT simulation with scripts for easy running, benchmark output post-processing and report generation. 

See the README for details!

README.md
hariharan-devarajan: I was thinking in these lines.
- we can have a base reader which is actual calls to npz.
- If we want to change how its implemented on pytorch or tf, we can extend it and do the extensions appropriately. 

This way we have a cleaner abstraction. What do you think?

src/reader/data_loader_reader.py
hariharan-devarajan: I was thinking in these lines.

we can have a base reader which is actual calls to npz.
If we want to change how its implemented on pytorch or tf, we can extend it and do the extensions appropriately.
This way we have a cleaner abstraction. What do you think?

requirements.txt
hariharan-devarajan: - Can we add the docker instructions also in the readme?
- Can we document it in the installation in a note that we need to install horovod manually.

src/reader/reader_handler.py
hariharan-devarajan: Maybe add a issue for this as enhancement.

src/reader/tf_reader.py
hariharan-devarajan: Maybe add a issue for this as enhancement.

requirements.txt
jovonho: - Done - see latest pushed changes! 
- Manually was the wrong word, it's just installed from the Dockerfile to be able to do some stuff with ldconfig around it.

zhenghh04:
Overall, I found no significant issues, except there are some small changes need to be made here and there. Because of the time, I'll merge this PR, and then we can fix those small issues later on. 

One big thing is to refactor the structure as "--framework ... --data-loader ... --format" to have data-loader on top of format to allow loading different formats through data loader. I have an implementation. I'll commit my changes after merging the PR.

PR #6, opened 2022-11-10
Some fixes 

jovonho:
Touch up README 
- fix typos, add details for report generation, tell user to flush caches between datagen and run.

Change default UNET3D and BERT params 
- I think we shouldn't encourage running datagen and run right after since the report generated will be wrong due to not flushing the caches. 
- Some of the parameter values were wrong e.g. file size for UNET3D

Fix iostat running 
- Add parameter to specify which io devices to trace 
- Change logging time to local time, to match with iostat (though this might not work on all machines depending on the time config)
- Since we're trying to remain general, you may not want to run iostat all the time? I dont know.

Add test configs for bert and unet that finish quickly to test the end-to-end process.

ISSUES:
- The BERT workload should currenty be run with only 1 process (mpirun -np 1) and the "num gpus" should be used as input to the number of computation threads and to scale the batch size (6 * num_gpus). Running bert with more than one process doesn't work currently and does not properly scale like the real workload (which uses a single process). We should specify this in the README, and add a parameter, though it will be redundant for UNET3D which simply uses the mpi num-procs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ output/
 notes/
 stuff/
 *.un~
+hydra_log/
+
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Deep Learining I/O (DLIO) Benchmark
+# Deep Learning I/O (DLIO) Benchmark
 
 This README provides a abbreviated documentation of the DLIO code. Please refer to https://argonne-lcf.github.io/dlio_benchmark/ for full user documentation. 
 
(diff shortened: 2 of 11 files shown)


zhenghh04:
@jovonho Could you take a look at my comments?

jovonho:
> @jovonho Could you take a look at my comments?

I don't see them!

zhenghh04:
For some reason, maybe the comments are only visible to myself. I quoted the comments below: 

* Regarding iostat: 

  Earlier today, I created another section in the yaml configure file, not knowing you also were trying to address this iostat issue.
  
  workflow: 
    train: True
    profiling: True
    
  profiling: 
    profiler: iostat
    iostat_command: .....
  In this way, one can specify the iostat command. We can also change it to iostat_devices, instead of specifying iostat_command.
  
  What I found is that the command might be different from system to system. Allowing users to specify command gives them more freedom. What do you think?
  
* Regarding yaml test files that you added: 

  The difference between unet3d.yaml and unet3d_test.yaml are the dataset size?
  
  I am afraid this will confuse people as which unet3d yaml files to use.
  
  How about performing all the test runs simply by overriding the configurations in the command line, instead of creating another file?
  
  You can check the GitHub action tests I created based on unet3d.yaml

=============================
I merged the pull request. For iostat profiler part, I use iostat_command, instead of adding devices. But we can discuss if you feel otherwise.

PR #7, opened 2022-11-13
PR3

jovonho:
Changed a few things:

Re-added the devices to trace as a parameter vs the iostat command:
- We don't want to expose any of the iostat command parameters to the user as they are all necessary to generate an output that the postprocessor is able to parse. We don't want the user to change anything EXCEPT the devices to trace, so it's logical to only expose that part.
- Added a line in README on how to install it since it is a dependency to generate the benchmark report.

Sharded the TFRecordDataset in tf_reader:
- The previous version was wrong and caused BERT with > 1 worker to hang forever. Now it's in line with the other tf.data loaders and works with multiple workers.

Changed how the number of steps is calculated and print the step number each iteration:
- Using ceiling to calculate the step number did not match reality for unet3d and bert. I tried with resnet50 as well, and it matched the expected. 

Added back and hid the test configs in `tests/`. It's just for convenience when testing so I don't have to change the real configs. Can use them like so 
```
mpirun -np 7 python3 src/dlio_benchmark.py --config-path $(pwd)/tests/configs workload=bert_test
```

Removed old scripts under `tests/` that are no longer relevant.

Removed TF and PT logging complaining about CUDA drivers / libraries


zhenghh04:
I merged this PR. Make a small change: io_devices_to_trace -> iostat_device. 

I still feel that if the test configs are just for yourself, it is OK to have it on your local machine, but not adding them to the github repo. People can always create their own custom test configs. What do you think?

jovonho:
Sure, that works fine! Thanks!

PR #8, opened 2022-11-22
Add docker image with CPU only dependencies

johnugeorge:
In this PR

1. Versions of packages are frozen in requirements file for reproducibility
2. CPU version of Dockerfile is added with no GPU dependencies

Docker file installation steps match the GitHub action DLIO installation


johnugeorge:
@jovonho

jovonho (approved):
Looks good to me!

zhenghh04:
@johnugeorge Are you able to check whether the dockerfile work with MacOSX? 

When I was running with bare metal version, there was some issue with iostat profiling in MacOSX system. Let us make sure this works before adding that to the dockerfile.

johnugeorge:
> @johnugeorge Are you able to check whether the dockerfile works in MacOSX?
> 
> When I was running with bare metal version, there was some issue with iostat profiling in MacOSX system. Let us make sure this works before adding that to the dockerfile.

@zhenghh04 Sorry, I didn't catch your comment on MacOSX.  IOStat is installed inside the debian docker container right? I assume, Github actions VM image already has IOStat installed in it for profiling. https://github.com/argonne-lcf/dlio_benchmark/blob/19e281623ee363feb454a6eb6fed5b0c4280ab14/.github/workflows/python-package-conda.yml#L7 

I was using MacOSX to build the image. I tried https://github.com/argonne-lcf/dlio_benchmark/blob/19e281623ee363feb454a6eb6fed5b0c4280ab14/.github/workflows/python-package-conda.yml#L46-L47 inside docker and it works. Anything else to be verified?

PR #20, opened 2022-12-05
Adding Storage interface for supporting multiple storage backends

johnugeorge:
This PR adds Storage interface in DLIO. This adds support for multiple storage backends like S3 objects etc 

**There is no change in current DLIO functionality or  behaviour with this PR** 

This PR adds

1. Adds one new config parameter in Dataset- `storage_type`. The default is `local_fs`.  The other option is `S3`
2. New Storage Interface is created for all filesystem operations 
3. For any benchmark, data dir and checkpoint dir has to be created on the storage system under test. While output folder is stored on the local client system itself.  So,  for all operations on data_dir and checkpoint_dir uses storage system interface.
4. Currently used local file system calls(using os module) are moved to file_storage.py. S3 apis are implemented in s3_storage.py
5. Changes are made only to TF framework as tf.io package can handle multiple system backends.  Support for Torch has to be added in a separate PR

---------------------------------

For @zhenghh04,  I am adding the design details as per discussion with @hariharan-devarajan 

A new storage interface is created which has generic concepts that can apply well for various storage backends. 
 
   Each storage has `namespace` with set of `nodes`. 
   `Namespace` is the mount point of the storage. For files, it is the root directory where input and checkpoint directories are created. For Objects, namespace refers to the bucket where input and checkpoint directories are created
   `Nodes` refer to the entries within the namespace which can be a file/directory/object. It has an `id` which is the `path` in Files and `Key` in `Objects`.

For a workload, the storage interface can use the framework IO apis for metadata and data operations  if framework native apis are available. eg: TF has `tf.io` which provides IO support for different storage backends. But Pytorch doesn't have it.   We will add PyTorch framework support for S3 later.

### DIFF
--- a/configs/workload/bert.yaml
+++ b/configs/workload/bert.yaml
@@ -9,7 +9,7 @@ workflow:
  checkpoint: True
  
 dataset: 
- data_folder: ./data/bert/
+ data_folder: data/bert
  format: tfrecord
  num_files_train: 500
  num_samples_per_file: 313532
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,7 +43,7 @@ tensorboard==2.11.0
 tensorboard-data-server==0.6.1
 tensorboard-plugin-wit==1.8.1
 tensorflow==2.11.0
-tensorflow-io==0.27.0
+tensorflow-io==0.28.0
 tensorflow-estimator==2.11.0
 termcolor==2.1.1
 torch==1.13.0
--- a/src/common/enumerations.py
+++ b/src/common/enumerations.py
@@ -17,6 +17,38 @@
 
 from enum import Enum
 
+class StorageType(Enum):
+    """
+    Different types of underlying storage
+    """
+    LOCAL_FS = 'local_fs'
+    PARALLEL_FS = 'parallel_fs'
+    S3 = 's3'
+
+    def __str__(self):
+        return self.value
+
+class MetadataType(Enum):
+    """
+    Different types of storage metadata
+    """
+    FILE = 'file'
+    DIRECTORY = 'directory'
+    S3_OBJECT = 's3_object'
+
+    def __str__(self):
+        return self.value
+
+class NamespaceType(Enum):
+    """
+    Different types of Storage Namespace
+    """
+    FLAT = 'flat'
+    HIERARCHICAL = 'Hierarchical'
+
+    def __str__(self):
+        return self.value
+
 class DatasetType(Enum):
     """
     Training and Validation
--- a/src/dlio_benchmark.py
+++ b/src/dlio_benchmark.py
@@ -17,7 +17,6 @@
 import os
 import math
 import hydra
-import shutil
 import logging
 import pandas as pd
 from time import time
@@ -38,10 +37,11 @@ from omegaconf import DictConfig, OmegaConf
 from src.utils.statscounter import StatsCounter
 from hydra.core.config_store import ConfigStore
 from src.utils.config import LoadConfig, ConfigArguments
-from src.common.enumerations import Profiler, DatasetType
+from src.common.enumerations import Profiler, DatasetType, StorageType
 from src.profiler.profiler_factory import ProfilerFactory
 from src.framework.framework_factory import FrameworkFactory
 from src.data_generator.generator_factory import GeneratorFactory
+from src.storage.storage_factory import StorageFactory
 
 
 class DLIOBenchmark(object):
@@ -77,7 +77,10 @@ class DLIOBenchmark(object):
         self.data_folder = self.args.data_folder
         self.output_folder = self.args.output_folder
         os.makedirs(self.output_folder, exist_ok=True)
-        os.makedirs(self.data_folder, exist_ok=True)
+        self.storage_root = self.args.storage_root
+        self.storage = StorageFactory().get_storage(self.args.storage_type, self.storage_root, self.args.framework)
+        if self.args.storage_root:
+            self.storage.create_namespace(exist_ok=True)
 
         self.framework = FrameworkFactory().get_framework(self.args.framework,
                                                           self.args.do_profiling)
@@ -344,8 +347,8 @@ class DLIOBenchmark(object):
                 logging.info(f"{utcnow()} Keep files set to False. Deleting dataset")
                 self.framework.barrier()
                 if self.my_rank == 0:
-                    if os.path.exists(self.args.data_folder):
-                        shutil.rmtree(self.args.data_folder)
+                    if self.storage.get_node(self.args.data_folder):
+                        self.storage.delete_node(self.args.data_folder)
                         logging.info(f"{utcnow()} Deleted data files")
             
             # Save collected stats to disk
--- a/src/framework/tf_framework.py
+++ b/src/framework/tf_framework.py
@@ -23,9 +23,11 @@ from src.common.error_code import ErrorCodes
 from src.framework.framework import Framework
 from src.reader.reader_factory import ReaderFactory
 from src.profiler.profiler_factory import ProfilerFactory
-from src.common.enumerations import FrameworkType, Profiler, FormatType, DatasetType
+from src.storage.storage_factory import StorageFactory
+from src.common.enumerations import FrameworkType, Profiler, FormatType, DatasetType, MetadataType
 
 import tensorflow as tf
+from tensorflow.python.framework import errors
 
 tf.compat.v1.logging.set_verbosity(tf.compat.v1.logging.ERROR)
 
@@ -48,6 +50,7 @@ class TFFramework(Framework):
     def init_reader(self, format_type, data_loader=None):
         self.reader_train = ReaderFactory.get_reader(format_type, data_loader=data_loader, dataset_type=DatasetType.TRAIN)
         self.reader_valid = ReaderFactory.get_reader(format_type, data_loader=data_loader, dataset_type=DatasetType.VALID)
+        self.storage = StorageFactory().get_storage(self.args.storage_type, self.args.storage_root, self.args.framework)
 
     def get_type(self):
         return FrameworkType.TENSORFLOW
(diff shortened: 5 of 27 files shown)


hariharan-devarajan:
@johnugeorge Can you please cleanup the commits with a rebase.
Have the following commit format.

Main subject
<new-line>
Details

src/framework/tf_framework.py
```diff
@@ -72,23 +74,23 @@ def checkpoint(self, epoch, step_number):
...
 
-            model_file = os.path.join(self.checkpoint_folder, f"model-{epoch}-{step_number}.bin")
-            meta_file = os.path.join(self.checkpoint_folder, f"meta-{epoch}-{step_number}.bin")
-            index_file = os.path.join(self.checkpoint_folder, f"index-{epoch}-{step_number}.bin")
+            model_file = self.storage.get_uri(os.path.join(self.checkpoint_folder, f"model-{epoch}-{step_number}.bin"))
+            meta_file = self.storage.get_uri(os.path.join(self.checkpoint_folder, f"meta-{epoch}-{step_number}.bin"))
+            index_file = self.storage.get_uri(os.path.join(self.checkpoint_folder, f"index-{epoch}-{step_number}.bin"))
 
-            f = open(model_file, "w")
+            f = tf.io.gfile.GFile(model_file, "w")
```
hariharan-devarajan, line 84: Even writing checkpoints should be abstracted through the storage interface. As in the absence of a filesystem we would have to write the checkpoints on S3 again.

src/framework/tf_framework.py
```diff
@@ -72,23 +74,23 @@ def checkpoint(self, epoch, step_number):
...
             string_val = "x" * self.args.model_size 
             f.write(string_val)
             f.close()
             # Should these scale with the model size?
-            f = open(index_file, "w")
+            f = tf.io.gfile.GFile(index_file, "w")
```
hariharan-devarajan, line 89: Abstract using storage interface

src/storage/storage_handler.py
```diff
@@ -0,0 +1,52 @@
...
+    def get_uri(self, file_name):
+        """
+            This method returns URI of a file name based on the implemented file system.
+            eg: For a file in S3, s3:// has to be prefixed to the file name.
+            eg: For a file in hdfs, hdfs:// has to be prefixed to the file name.
+        """
+        pass
+
+    @abstractmethod
+    def makedirs(self, dir_name, exist_ok=False):
```
hariharan-devarajan, line 19: we should name it more generic. The general term is a namespace. Even in filesystem, directories represent a namespace for a file. i would prefer using create_namespace() API.

src/storage/storage_handler.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+            This method creates a directory in the backend system.
+            If exist_ok flag is set, no error is returned if dir is already present.
+        """
+        pass
+
+    @abstractmethod
+    def listdir(self, dir_name):
```
hariharan-devarajan, line 27: This could be called walk()/traverse()/visit(). I prefer walk() as we can add flags for recursively walk if needed.

src/storage/storage_handler.py
```diff
@@ -0,0 +1,52 @@
+from abc import ABC, abstractmethod
+
+from src.utils.config import ConfigArguments
+
+class DataStorage(ABC):
```
hariharan-devarajan, line 5: General abstraction should 
DataStorage
- object dir/file/object - node
- path of file/directories - namespace
- file/objects - leaf-node

src/storage/storage_handler.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @abstractmethod
+    def exists(self, path):
+        """
+            This method checks if path already exists.
+        """
+        pass
+
+    @abstractmethod
+    def rmtree(self, dir_name):
```
hariharan-devarajan, line 41: we should call this remove with recurse as an optional arguments.

src/storage/storage_handler.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+            This method removes the directory recursively.
+        """
+        pass
+
+    @abstractmethod
+    def isfile(self, file_name):
```
hariharan-devarajan, line 48: Instead of checking for is file, we should get type of node and the client of this call can check the type of node as file/object/namespace

src/common/enumerations.py
```diff
@@ -17,6 +17,17 @@
 
 from enum import Enum
 
+class StorageType(Enum):
+    """
+    Different types of underlying storage
+    """
+    FILE = 'file'
```
hariharan-devarajan, line 24: FILE or object is a type
We should use FILESYSTEM and S3 as Storage Type. I would further prefer if we specialised it further as LOCAL_FILESYSTEM, PARALLEL_FILE_SYSTEM, and S3

hariharan-devarajan (changes requested):
Please check my inputs on the Storage abstractions.

zhenghh04:
@johnugeorge @hariharan-devarajan If it is OK with you all. I'll keep this PR open until we have everything related to S3 implemented before merging it to the main branch. Feel free to keep committing changes to this PR.

configs/workload/bert.yaml
```diff
@@ -28,4 +28,4 @@ reader:
 
 checkpoint:
  steps_between_checkpoints: 1250
- model_size: 4034713312
```
hariharan-devarajan, line 31: Revert this. I think it was done by mistake.

configs/workload/bert.yaml
```diff
@@ -9,7 +9,7 @@ workflow:
  checkpoint: True
  
 dataset: 
- data_folder: ./data/bert/
+ data_folder: data/bert
```
hariharan-devarajan, line 12: Can we by default infer the relative path. I see other files have similar chnages.

requirements.txt
```diff
@@ -43,7 +43,7 @@ tensorboard==2.11.0
 tensorboard-data-server==0.6.1
 tensorboard-plugin-wit==1.8.1
 tensorflow==2.11.0
-tensorflow-io==0.27.0
+tensorflow-io==0.28.0
```
hariharan-devarajan, line 46: Why did we increase the version requirement?

src/framework/tf_framework.py
```diff
@@ -44,6 +45,8 @@ def __init__(self, profiling):
     def init_reader(self, format_type, data_loader=None):
         self.reader_train = ReaderFactory.get_reader(format_type, data_loader=data_loader, dataset_type=DatasetType.TRAIN)
         self.reader_valid = ReaderFactory.get_reader(format_type, data_loader=data_loader, dataset_type=DatasetType.VALID)
+        self.storage = StorageFactory().get_storage(self.args.storage_type, self.args.storage_root,
+                                                                        self.args.framework)
```
hariharan-devarajan, line 49: Why do we have a whitespace here?

src/dlio_benchmark.py
```diff
@@ -77,7 +77,10 @@ def __init__(self, cfg):
         self.data_folder = self.args.data_folder
         self.output_folder = self.args.output_folder
         os.makedirs(self.output_folder, exist_ok=True)
-        os.makedirs(self.data_folder, exist_ok=True)
+        self.storage = StorageFactory().get_storage(self.args.storage_type, self.args.storage_root,
+                                                                        self.args.framework)
```
hariharan-devarajan, line 81: fix whitespace

src/framework/tf_framework.py
```diff
@@ -72,26 +75,21 @@ def checkpoint(self, epoch, step_number):
...
             model_file = os.path.join(self.checkpoint_folder, f"model-{epoch}-{step_number}.bin")
             meta_file = os.path.join(self.checkpoint_folder, f"meta-{epoch}-{step_number}.bin")
             index_file = os.path.join(self.checkpoint_folder, f"index-{epoch}-{step_number}.bin")
 
-            f = open(model_file, "w")
             string_val = "x" * self.args.model_size 
-            f.write(string_val)
-            f.close()
+            self.storage.put_data(model_file, string_val)
             # Should these scale with the model size?
```
hariharan-devarajan, line 87: Add a TODO so that we can get to it. ALso create a issue.

src/framework/torch_framework.py
```diff
@@ -57,6 +58,8 @@ def __init__(self, profiling):
     def init_reader(self, format_type, data_loader=None):
         self.reader_train = ReaderFactory.get_reader(format_type, data_loader=data_loader, dataset_type=DatasetType.TRAIN)
         self.reader_valid = ReaderFactory.get_reader(format_type, data_loader=data_loader, dataset_type=DatasetType.VALID)
+        self.storage = StorageFactory().get_storage(self.args.storage_type, self.args.storage_root,
+                                                                                self.args.framework)
```
hariharan-devarajan, line 62: whitespace.

src/reader/reader_handler.py
```diff
@@ -63,15 +64,18 @@ def __init__(self, dataset_type):
         self.dataset_type = dataset_type
         self.framework = FrameworkFactory().get_framework(self._args.framework,
                                                           self._args.do_profiling)
+        self.storage = StorageFactory().get_storage(self._args.storage_type, self._args.storage_root,
+                                                                        self._args.framework)
+        # We do this here so we keep the same evaluation files every epoch
```
hariharan-devarajan, line 69: Duplicated comment.

src/reader/reader_handler.py
```diff
@@ -63,15 +64,18 @@ def __init__(self, dataset_type):
         self.dataset_type = dataset_type
         self.framework = FrameworkFactory().get_framework(self._args.framework,
                                                           self._args.do_profiling)
+        self.storage = StorageFactory().get_storage(self._args.storage_type, self._args.storage_root,
+                                                                        self._args.framework)
```
hariharan-devarajan, line 68: whitespace.

src/storage/file_storage.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def walk_node(self, id):
+        return os.listdir(self.get_uri(id))
+
+    def delete_node(self, id):
+        shutil.rmtree(self.get_uri(id))
+        return True
+
+    def put_data(self, id, data, offset=None, length=None):
+        with open(self.get_uri(id), "w") as fd:
```
hariharan-devarajan, line 49: Add a todo for partial read and write.

hariharan-devarajan (changes requested):
Overall it looks great.

hariharan-devarajan:
Can you check why the build is failing and fix that as well.

requirements.txt
johnugeorge: Currently, there is a warning during `pip install -r requirements.txt` stating the incompatibility with the current tensorflow version. `tensorflow-io` needs to be 0.28.0 for the right match

From logs,
`ERROR: tensorflow-io 0.27.0 has requirement tensorflow-io-gcs-filesystem==0.27.0, but you'll have tensorflow-io-gcs-filesystem 0.28.0 which is incompatible.`

configs/workload/bert.yaml
johnugeorge: The field is not removed.  The last line is not currently newline terminated.  https://github.com/argonne-lcf/dlio_benchmark/pull/20/files#diff-a71b531b049497d6eca434150d4abf41b058d3956ac2d3c715abd0e291d41ab9R31

johnugeorge:
All changes added @hariharan-devarajan @zhenghh04

hariharan-devarajan (approved):
I think it is ready for merging.

johnugeorge:
Thanks @hariharan-devarajan

PR #22, opened 2022-12-07
Minor fixes 

zhenghh04:
In this PR, I fixed a bunch of minor issues: 

1) changed copy right symbol to (c) in case it causes trouble in some os. 
2) changed "data_reader" to "reader" in YAML config to be consistent with the naming in the code. 
3) set progress bar to stdout for generating the dataset. 
4) added timing decorator for dataset loading. Detailed timing can be done in debug mode. 
5) removed "file_access" parameter. This will be determined automatically through num_files_train and num_samples_per_file
6) removed unit test for "iostat" profiling in tests/dlio_benchmark_test.py. It causes hang when run in MPI. But this test was able to run outside of pytest. Therefore, I removed it for now, and will investigate further why pytest-mpi cause this test to hang.

### DIFF
--- a/src/reader/torch_data_loader_reader.py
+++ b/src/reader/torch_data_loader_reader.py
@@ -1,5 +1,5 @@
 """
-   Copyright © 2022, UChicago Argonne, LLC
+   Copyright (c) 2022, UChicago Argonne, LLC
    All Rights Reserved
 
    Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/reader/torch_data_loader_reader.py
+++ b/src/reader/torch_data_loader_reader.py
@@ -20,7 +20,7 @@ import numpy as np
 from time import time
 import os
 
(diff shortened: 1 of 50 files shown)


src/reader/torch_data_loader_reader.py
```diff
@@ -97,9 +97,10 @@ def read(self, epoch_number):
         # superclass function shuffle the file list 
         super().read(epoch_number)
         do_shuffle = True if self.memory_shuffle != Shuffle.OFF else False
-        
-        dataset = TorchDataset(self._file_list, self.my_rank, self.format)
-        
+        if self._args.num_samples_per_file == 1:
+            dataset = TorchDataset(self._file_list, self.my_rank, self.format)
```
hariharan-devarajan, line 101: We should create a issue to support multiple files for PyTorch so that we can pick it up later.

PR #23, opened 2022-12-07
Add ckpting to UNET3D workload, remove old prefetch param

jovonho:
- Add checkpointing after every eval for UNET3D
- Remove old prefetch param & align hdf5_stimulate (btw, should it be called simulate?) that was the only one still using it
- add checkpoints folder to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Benchmark generated data
 data/
 output/
+checkpoints/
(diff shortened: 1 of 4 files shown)


zhenghh04 (approved):
For the prefetch, 

I did the same thing in my PR also. The option is removed.

PR #48, opened 2023-03-18
Trace improvement.

hariharan-devarajan:
- Added logging to all functions.
- categories are correctly set so that they can be filtered.
- format of JSON is corrected.
- Additional args to events of internal profile to support filtering. 
  - step_number, epoch_number, image_global_index, image_size


hariharan-devarajan:
Added sample [notebook](https://github.com/hariharan-devarajan/notebooks/blob/dev/Untitled.ipynb) that shows how the data will be parsed.

PR #57, opened 2023-03-30
Merging branch mlperf_storage_v0.5.1 

zhenghh04:
Merging https://github.com/argonne-lcf/dlio_benchmark/pull/56
- summary.json is included in the output folder which includes the metric info
- 0_output.json, 1_output.json, etc contain the per process output info
- AUU and throughput are calculated by averaging / summing all the processes.
- Added read bandwidth estimated by throughput * record_length (does not take into account record_length_stdev)


src/dlio_benchmark.py
```diff
@@ -232,49 +207,23 @@ def _eval(self, epoch):
...
-            t1 = time()
-            time_epoch['time_per_step'].append(t1 - t0)
-            self.stats.eval_batch_processed(epoch, step, t0)
+            self.stats.eval_batch_processed(epoch, step, t0, eval_time)
 
             step += 1
             if step > total:
                 break
             t0 = time()
-            self.framework.barrier()
```
hariharan-devarajan, line 264: Did we remove barrier per step?

src/dlio_postprocessor.py
```diff
@@ -49,17 +50,17 @@ def __init__(self, args) -> None:
         self.verify_and_load_all_files()
         self.disks = []
         self.overall_stats = {}
-
+        self.record_size = args.record_size
```
hariharan-devarajan, line 53: Why do we keep copying args into every class. It increases the memory usage unnecessary. Please use refer to args singleton and use that directly.

src/dlio_postprocessor.py
```diff
@@ -190,9 +193,9 @@ def process_loading_and_processing_times(self):
         logging.info(f"Computing overall stats")
 
         # Save the overall stats
-        self.overall_stats['samples/s'] = self.get_stats(all_samples_per_s)
-        self.overall_stats['sample_latency'] = self.get_stats(all_sample_latencies)
-
+        self.overall_stats['samples/s'] = self.get_stats(self.summary['metric']['train_throughput_samples_per_second'])
```
hariharan-devarajan, line 196: This feels very verbose. I prefer the apis to be simpler. Just get the data u want and then use that value next.

tests/test_data/0_output.json
```diff
@@ -0,0 +1,169 @@
+{
+    "hostname": "7a3725255f7c",
+    "1": {
+        "load": {
+            "block1": [
```
hariharan-devarajan, line 5: What is block? is it step? I prefer we use same terminology here. So that it's intuitive.

hariharan-devarajan (changes requested):

src/dlio_postprocessor.py
zhenghh04: Because args is only available in __init__

src/dlio_benchmark.py
zhenghh04: No. Barrier per step is needed. Each time step, allreduce is called in actual training

src/dlio_postprocessor.py
hariharan-devarajan: I mean we should dont need to pass the args at all, we can get the singleton here and use that directly as a class variable.

hariharan-devarajan (approved):
In general, I dont see any major issues. Left some minor comments.

tests/test_data/0_output.json
zhenghh04: Yes, this came up by Loic. Block composed by a set of training steps that do not contain check pointing. 

I don't have another word to replace that yet. So I just keep whatever it is now.

PR #66, opened 2023-05-27
For X86 and AMD machines, we can create a pip based dlio installations

hariharan-devarajan:
Building an installable package needs the following features

- Correct source folder name. It should not be src but the project name.
- Then, the data files being part of the package is recommended. E.g., configs.
- adding namespace files i.e., __init__.py, in all module files. This will help set up files to copy the correct namespaces and module files.
- Fix the main benchmark module from dlio_benchmark.py to benchmark.py.

With these changes, users can do ```pip install .``` to install the dependencies and main dlio_benchmark as an installation script and copy it to the bin. Allowing users to execute the benchmark using ```dlio_benchmark -h```.


hariharan-devarajan:
> Any particular reason you change the folder name from src to dlio_benchmark?

Yes, when we install using setup.py, the package which was earlier src was getting copied into the site-packages. I was looking at other python packages such as hydra and noticed their top level is called the same as the package name for the reason of installation purposes. That is why I change it that way.

hariharan-devarajan:
@zhenghh04 

I fixed the following.
1. refactored benchmark and postprocesser script.
2. fixed documentation.

zhenghh04:
> @zhenghh04
> 
> I fixed the following.
> 
> 1. refactored benchmark and postprocesser script.
> 2. fixed documentation.

I saw some extra files are been added to the repo, such as ./build, ./hydra_log. Could you clean that up?

hariharan-devarajan:
@zhenghh04 Fixed.

zhenghh04 (approved):

Issue #69, opened 2023-06-15
Impact of affinity on read threads in DLIO.


hariharan-devarajan:
In Livermore Computing machines, srun or lrun commands set a core affinity of 1 core which could make all threads on the node run using 1 core even if more cores are available. 

We should increase the core affinity on our mpirun to make use of all cores in the system. If users have an affinity less than read threads, we should throw a warning to let them know.

hariharan-devarajan:
@zhenghh04 What do u think of this?

PR #72, opened 2023-06-20
Added validation for checking the core affinity of the process.

hariharan-devarajan:
This validation does not create an error but only a warning message for the users.
It will be thrown when users have set read_threads < core affinity of the process.


hariharan-devarajan:
fixes #69

PR #82, opened 2023-08-18
Fixing subfolder issues and added subset tests

zhenghh04:


docs/source/examples.rst
```diff
@@ -12,7 +12,7 @@ UNET3D: 3D Medical Image Segmentation
 
 .. code-block:: yaml
 
-    # contents of unet3d.yaml
+    # contents of unßet3d.yaml
```
hariharan-devarajan, line 15: Non ASCII character introduced.

PR #96, opened 2023-10-06
fixed resnet50 config and sleep issue

zhenghh04:


zhenghh04:
This is to address the issue: https://github.com/argonne-lcf/dlio_benchmark/issues/95

PR #110, opened 2023-11-08
Add new FSSPEC storage type

krehm:
Add a new FSSPEC storage type implemented using the fsspec filesystem specification.  This is in preparation for adding a new DAOS fsspec plugin so that dlio_benchmark can be used directly with the DAOS DFS library. fsspec also makes it possible to use S3 with both TF and PyTorch.  Other fsspec filesystem emulations like hdf5 are also possible.


krehm:
Accidental pull request, sorry.

PR #129, opened 2023-12-16
Add support for pytorch spawn and forkserver multiprocessing_context

krehm:
…values

The ConfigArguments class singleton must be pickled by the TorchDataset class __init__ method so that the main process's current configuration can be then reconsituted within a spawn or forkserver child processes by method worker_init.


dlio_benchmark/data_loader/torch_data_loader.py
```diff
@@ -44,11 +46,14 @@ def __init__(self, format_type, dataset_type, epoch, num_samples, num_workers, b
...
         self.batch_size = batch_size
+        args = ConfigArguments.get_instance()
+        self.serial_args = pickle.dumps(args)
         if num_workers == 0:
             self.worker_init(-1)
 
     @dlp.log
     def worker_init(self, worker_id):
+        pickle.loads(self.serial_args)
         logging.debug(f"{utcnow()} worker initialized {worker_id} with format {self.format_type}")
```
hariharan-devarajan, line 57: How would this update the singleton? Won’t we have to have a static method in config to make sure we update the config with this loaded instance on the init of worker. 

Also can u add a new torch_spawn_data_loader.py class which add the multiprocessing context to the data loader as spawn and that to tests to make sure it all works.

hariharan-devarajan (changes requested):
Thank you for giving this a try. I have a few suggestion/questions on this.

My understanding for this is as follows.
1. Create a new data loader called torch_spawn_data_loader.py
2. Set multiprocessing context as spawn in data loader call.
3. As u have it now, pickle the state of configuration.
4. Load the pickle and have an update method which updates all values from this pickled object. The set state method should unpickle but i am unsure if it would correctly update our already initialized singleton. If this works without an update method then it’s fine. 
5. Add pytest to add data loader for all tests as needed.
6. Add these tests to CI in the correct sections.

Let me know if this makes sense.

dlio_benchmark/data_loader/torch_data_loader.py
krehm: When pickle.loads() is called, pickle sees that the ConfigArguments class now has a __setstate__ method.  That causes pickle to populate the memory-allocated class object with all the fields in the pickle, but it does NOT call the ConfigArguements __init__ method.  The __setstate__ method then sets ConfigArguments.__instance to the newly created/unpickled ConfigArguments class, so any code which references ConfigArguments will now always get the new unpickled class.

krehm: I am working on tests.

krehm:
A new pytorch loader is not necessary, the existing one supports multiprocessing_context.

dlio_benchmark/data_loader/torch_data_loader.py
zhenghh04: @krehm I added the tests to github actions. Let us see whether it works.

zhenghh04: @krehm @hariharan-devarajan looks like it does not work. Please see the github action runs.

krehm:
It appears there are issues mixing python multiprocessing with MPI.   I will take a look.

krehm:
The gist of the problem is that because dlio_benchmark uses mpi4py, there is an open socket file descriptor called PMI_fd which is used to communicate with the mpich hydra process manager, for example to retrieve one's rank and comm_size.  When a spawn or forkserver multiprocessing_context is used, the resulting child process doesn't have that file descriptor because it is not a straight fork of the parent process, and so when the child tries to talk to hydra, it fails because the fd that it thinks is connected to hydra is not.   I'm surprised that it even works in the fork case; when read_threads=4 you have four additional child processes in addition to their parent all talking to hydra on the same socket and claiming to be the same MPI rank.   That seems really racy at best.   In all of these cases these child processes really should not be using MPI, they are not valid ranks in the mpirun job, they are just worker processes.  Anyway, I am still searching for a solution.

krehm:
OK, I see how to fix this, should have something by tomorrow.

dlio_benchmark/utils/utility.py
```diff
@@ -51,12 +56,52 @@ def utcnow(format=LOG_TS_FORMAT):
     return datetime.now().strftime(format)
 
 
-def get_rank():
-    return MPI.COMM_WORLD.rank
-
-
-def get_size():
-    return MPI.COMM_WORLD.size
+class mpi:
```
hariharan-devarajan, line 59: Change name to DLIOMPI so that we don’t conflict classes with others and maintain correct formatting

dlio_benchmark/utils/utility.py
```diff
@@ -51,12 +56,52 @@ def utcnow(format=LOG_TS_FORMAT):
...
+    __instance = None
+
+    def __init__(self):
+        if mpi.__instance is not None:
+            raise Exception(f"Class {self.__init__.__qualname__} is a singleton!")
+        else:
+            self.mpi_rank = -1
```
hariharan-devarajan, line 66: Initialize rank to 0 always. Otherwise some implementations will fail.

dlio_benchmark/utils/utility.py
```diff
@@ -51,12 +56,52 @@ def utcnow(format=LOG_TS_FORMAT):
...
+            self.mpi_size = -1
```
hariharan-devarajan, line 67: Initialize this to 1 by default.

hariharan-devarajan (changes requested):
Some minor comments.

dlio_benchmark/utils/utility.py
krehm: Yes, this it better.

dlio_benchmark/utils/utility.py
krehm: These two cannot change, they are necessary for detecting a process which tries to use its rank or comm_size before DLIOMPI.initialize() has been called (in the benchmark parent process case) or before the parent's rank and comm_size have been initialized by DLIOMPI.set_parent_values() in a pytorch read_thread child process.  See the checks in rank(), size(), and comm().  read_threads children created via 'fork' are especially tricky, they appear to have been MPI-initialized because they are a fork, but they need to be prevented from making calls like barrier() because they are not members of the MPI job, they don't have their own rank values, and comm_size doesn't include them.   They can however use their parent's rank and comm_size values for math related to deciding which samples to read or whatever.

hariharan-devarajan: I see. But wouldn't having an explicit variable to detect this would be better. Repurposing rank and comm size for this could lead to create bugs in the future. 

I would recommend adding a variable which can explicitly track this.

krehm: Yes, I was thinking about this as well.  Having child processes using the same rank() and comm() functions is confusing.  It would be better if they were clearly retrieving their parent's values.  I'll work on this.

krehm:
Should I add the test_pytorch_multiprocessing_context tests to python-package-conda.yml as part of this PR?   I have run each of those tests by hand, and they all pass.

zhenghh04:
> Should I add the test_pytorch_multiprocessing_context tests to python-package-conda.yml as part of this PR? I have run each of those tests by hand, and they all pass.

Yes. Please add the tests to python-package-conda.yml

dlio_benchmark/utils/utility.py
```diff
@@ -51,12 +57,83 @@ def utcnow(format=LOG_TS_FORMAT):
...
-
-
-def get_size():
-    return MPI.COMM_WORLD.size
+# After the DLIOMPI singleton has been instantiated, the next call must be
+# either initialize() if in an MPI process, or set_parent_values() if in a
+# non-MPI pytorch read_threads child process.
+class DLIOMPI:
+    __instance = None
+    class MPI_state(Enum):
```
hariharan-devarajan, line 65: Following code convention in the repo, lets use MPIState and you can add it to the enumerations.

hariharan-devarajan (changes requested):
One last small fix.

dlio_benchmark/utils/utility.py
```diff
@@ -51,12 +57,83 @@ def utcnow(format=LOG_TS_FORMAT):
...
+        uninitialized = 1
```
hariharan-devarajan, line 66: Also, check enumerations.py to move this there and call the enumerations

UNINITIALIZED = 0
MPI_INITIALIZED = 1 
CHILD_INITIALIZED = 2

dlio_benchmark/utils/utility.py
krehm: Are you saying that you want it added to enumerations.py?   Because the enum has no meaning outside of the DLIOMPI class, it is completely internal to the class.

hariharan-devarajan (changes requested):

dlio_benchmark/utils/utility.py
hariharan-devarajan: I know, for now we don’t have any meaning. But eventually if we implement other communication premitives, its would be better. It’s just a code convention, we have been following.

hariharan-devarajan (approved):
@krehm @zhenghh04 The changes look good.

Thank you for all your effort.

zhenghh04:
> @krehm @zhenghh04 The changes look good.
> 
> Thank you for all your effort.

Thank you for the effort. I'll merge it to the main branch.

PR #137, opened 2024-01-10
Support for custom checkpointing.

hariharan-devarajan:
This feature is required for us to try different checkpointing strategies in the future.


hariharan-devarajan:
@zhenghh04 Can we work on merging PR.

zhenghh04 (approved):

PR #142, opened 2024-01-24
Fixed printing for DLIO output.

hariharan-devarajan:
Currently, we print the dlp logger file even if DLIO is disabled. Also, we need to print only for rank 0—similar changes for all factories.


zhenghh04 (approved):

hariharan-devarajan:
@zhenghh04 Can we merge this?

zhenghh04 (approved):

PR #156, opened 2024-02-26
Optimization to sample distribution

TheAssembler1:
Removes the redundant calls to `os.path.abspath`


hariharan-devarajan (approved):
This looks good.

PR #194, opened 2024-05-28
Improve tfreader parsing performance (batch)

LouisDDN:
Instead of processing the elements one-by-one in the reader pipeline, we use a batch. 

After having applied this modification, perf shows that a couple of functions take less time. 
These function are, I believe , related to C function call switches and python interpreter:
EagerTensor_init took 12% of the time => now 0.26%
pybind11::cpp_function::dispatcher took 10% => now 0.64%
PyEval_EvalFrameDefault took 9% => now 0.19% 

On the same test system, it doubles the throughput of resnet50 (1x H100).
This throughput was also measured on the network:

Test | Throughput (MB/s) | AU (%) | # reader threads
-- | -- | -- | --
Without   patch, 1x H100 | 115.6499 | 59.3728 | 4
With   patch, 1x H100 | 193.4854 | 99.417 | 4

### DIFF
--- a/dlio_benchmark/reader/tf_reader.py
+++ b/dlio_benchmark/reader/tf_reader.py
@@ -86,21 +86,19 @@ class TFReader(FormatReader):
         self._dataset = tf.data.TFRecordDataset(filenames=self._file_list, buffer_size=self._args.transfer_size, 
                                                 num_parallel_reads=self._args.read_threads)
 
-        self._dataset = self._dataset.batch(self.batch_size)
-        self._dataset = self._dataset.map(
-                lambda x: tf.py_function(func=self._parse_image, inp=[x], Tout=[tf.uint8]), 
-                num_parallel_calls=self._args.computation_threads)
-        self._dataset = self._dataset.unbatch()
-
-        self._dataset = self._dataset.shard(num_shards=self._args.comm_size, index=self._args.my_rank)
-
         if self._args.sample_shuffle != Shuffle.OFF:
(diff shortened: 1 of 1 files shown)


LouisDDN:
New pipeline order is:
- shuffle
- shard
- batch
- map
- repeat

Previous was:
- map
- shard
- shuffle
- batch
- repeat 

I believe this new order is correct but I would appreciate additional reviews.

LouisDDN:
Here is additional result. 
The performance boost is quite dramatic.

image

The patch applied to have "no compute time"
```diff
diff --git a/dlio_benchmark/main.py b/dlio_benchmark/main.py
index ebeb5a9..b4cdad4 100644
--- a/dlio_benchmark/main.py
+++ b/dlio_benchmark/main.py
@@ -269,7 +269,7 @@ class DLIOBenchmark(object):
                     computation_time = random.normal(self.computation_time, self.computation_time_stdev)
                 else:
                     computation_time = self.computation_time
-            self.framework.compute(batch, epoch, block_step, computation_time)
+            # self.framework.compute(batch, epoch, block_step, computation_time)
             self.stats.batch_processed(epoch, overall_step, block, t0, computation_time)
             self.comm.barrier()
             if self.do_checkpoint and (
```

zhenghh04 (approved):

zhenghh04:
Thanks @LouisDDN, this new PR looks good to me. I'll merge that after al the tests are passed.

PR #198, opened 2024-06-04
Fix # of files not divisible by # of accelerators

LouisDDN:
I believe there is one MPI process that exit the training "for" loop. 
It incorrectly synchronize using a different MPI_Barrier. As as result, all but one processes are stuck in the next MPI_Barrier. 

To fix, reduce the # of files.

### DIFF
--- a/dlio_benchmark/main.py
+++ b/dlio_benchmark/main.py
@@ -196,6 +196,14 @@ class DLIOBenchmark(object):
                 file_list_train = fullpaths
             elif dataset_type is DatasetType.VALID:
                 file_list_eval = fullpaths
+        if not self.generate_only and self.num_files_train > self.comm_size and self.num_files_train % self.comm_size != 0:
+            self.num_files_train = self.num_files_train - (self.num_files_train % self.comm_size)
+            logging.warning(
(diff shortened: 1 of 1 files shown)


LouisDDN:
Using gdb with 4 simulated H100, I see 1 MPI process stuck in MPI_Reduce & all of the other in MPI_Barrier which seems to confirm my first comment.

hariharan-devarajan:
The only way this is handled currently in some AI workloads are by doing a all reduce with MIN aggregation. @zhenghh04 What do u think.

zhenghh04:
@LouisDDN We need an alternative solution here. Reducing the number of files is not so good. Will create issue for other cases.

LouisDDN:
Can we add a parameter in the yaml file like "autosplit_number_of_files" that would be set to one for MLPerf & 0 by default? 
Would that work like this?

zhenghh04:
@LouisDDN Please check an alternative way: #200. It is essentially equivalent at the end for most of the cases, but for the case, num_files < comm.size. #200 will work.

zhenghh04:
I'll close this one PR now, since an alternative solution has been merged.

PR #213, opened 2024-07-23
Bug fix for MPI and HDF5 workloads on LC clusters

hariharan-devarajan:
This PR fixes MPI and HDF5 workloads in LC machine.

- [X] Fixed MPI initializes and finalization
- [X] Fix the data generation for chunked HDF5


hariharan-devarajan:
I will do a different PR.

PR #216, opened 2024-07-29
Fixed iterator to only store data for that rank.

hariharan-devarajan:
- we fix reconfiguration logic to build rank map for train and valid.
- Now we need reconfigure per epoch to shuffle the list.


zhenghh04 (approved):
This PR addresses reconfiguration. 

In reconfigure, global shuffling is performed. The data loader only have the indices of the local samples. It still has the shuffling on the pytorch / dali data loader which need to be fixed. 

Please send another PR to fixed the redundant shuffling issue @hariharan-devarajan

PR #222, opened 2024-08-27
Support for setting different DLIO_LOG_LEVEL

zhenghh04:
In this PR, we changed the per step output from info to debug to reduce the logging overhead. 

We added support for changing logging level through environment variable `DLIO_LOG_LEVEL` which can be set to be error, warning, output, info, debug. By default we set as output. 

All the per-step log is put >= info level. The output level only contains per epoch logs. With this, the log file size is significantly reduced. 

In the debug level, the log format contains source file and line number, while other levels do not.

### DIFF
--- a/dlio_benchmark/common/enumerations.py
+++ b/dlio_benchmark/common/enumerations.py
@@ -54,9 +54,10 @@ class LogLevel(Enum):
     """
     Different levels of logging
     """
-    DEBUG = "debug"
-    INFO = "info"
+    ERROR = "error" 
     WARN = "warn"
+    INFO = "info"
+    DEBUG = "debug"
     def __str__(self):
         return self.value
 
@@ -115,7 +116,6 @@ class ComputationType(Enum):
     NONE = 'none'
     SYNC = 'sync'
     ASYNC = 'async'
-
 
 class FormatType(Enum):
     """
--- a/dlio_benchmark/common/enumerations.py
+++ b/dlio_benchmark/common/enumerations.py
@@ -54,10 +54,10 @@ class LogLevel(Enum):
     """
     Different levels of logging
     """
-    ERROR = "error" 
-    WARN = "warn"
-    INFO = "info"
     DEBUG = "debug"
+    INFO = "info"
+    WARNING = "warning"
+    ERROR = "error" 
     def __str__(self):
         return self.value
 
--- a/dlio_benchmark/common/enumerations.py
+++ b/dlio_benchmark/common/enumerations.py
@@ -50,6 +50,15 @@ class StorageType(Enum):
 
     def __str__(self):
         return self.value
+class LogLevel(Enum):
+    """
+    Different levels of logging
+    """
+    DEBUG = "debug"
+    INFO = "info"
+    WARN = "warn"
+    def __str__(self):
+        return self.value
 
 class MetadataType(Enum):
     """
--- a/dlio_benchmark/common/enumerations.py
+++ b/dlio_benchmark/common/enumerations.py
@@ -50,16 +50,6 @@ class StorageType(Enum):
 
     def __str__(self):
         return self.value
-class LogLevel(Enum):
-    """
-    Different levels of logging
-    """
-    DEBUG = "debug"
-    INFO = "info"
-    WARNING = "warning"
-    ERROR = "error" 
-    def __str__(self):
-        return self.value
 
 class MetadataType(Enum):
     """
--- a/dlio_benchmark/utils/utility.py
+++ b/dlio_benchmark/utils/utility.py
@@ -56,7 +56,7 @@ except:
         def reset(self):
             return
         def log_static(self, func):
-            return
+            return func
     class dftracer(object):
         def __init__(self,):
             self.type = None
--- a/dlio_benchmark/utils/utility.py
+++ b/dlio_benchmark/utils/utility.py
@@ -40,11 +40,11 @@ except:
         def __init__(self,  cat, name=None, epoch=None, step=None, image_idx=None, image_size=None):
             return 
         def log(self,  func):
-            return 
+            return func
         def log_init(self,  func):
-            return 
+            return func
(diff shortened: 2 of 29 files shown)


dlio_benchmark/common/enumerations.py
```diff
@@ -50,6 +50,15 @@ class StorageType(Enum):
 
     def __str__(self):
         return self.value
+class LogLevel(Enum):
+    """
+    Different levels of logging
+    """
+    DEBUG = "debug"
+    INFO = "info"
+    WARN = "warn"
```
hariharan-devarajan, line 59:
```suggestion
    WARN = "warn"
    ERROR = "error"
```

hariharan-devarajan (changes requested):
We should add error logging as an option for pure performance mode. 
For the final per metric we should use a different logging, which outputs PRINT messages to STDOUT and redirects to the dlio.log file. 

The other logging is optional from the benchmark.

hariharan-devarajan:
@zhenghh04 Can we not print the file name and line number for info logging. We should only do that for debug logging. This will significantly reduce log size.

zhenghh04:
@hariharan-devarajan , made the changes as you suggested. Please review it again.

dlio_benchmark/utils/utility.py
```diff
@@ -37,38 +37,38 @@
...
 except:
     class Profile(object):
-        def __init__(self,  **kwargs):
-            return 
-        def log(self,  **kwargs):
-            return 
-        def log_init(self,  **kwargs):
-            return 
-        def iter(self,  **kwargs):
+        def __init__(self,  cat, name=None, epoch=None, step=None, image_idx=None, image_size=None):
```
hariharan-devarajan, line 40: We should be reverting these changes.

dlio_benchmark/utils/utility.py
```diff
@@ -37,38 +37,38 @@
...
+            return func
+        def iter(self,  func, iter_name="step"):
+            return func
         def __enter__(self):
             return
-        def __exit__(self, **kwargs):
+        def __exit__(self, type, value, traceback):
             return
-        def update(self, **kwargs):
+        def update(self, epoch=None, step=None, image_idx=None, image_size=None, args={}):
```
hariharan-devarajan, line 52: Lets revert this

dlio_benchmark/utils/utility.py
```diff
@@ -37,38 +37,38 @@
...
+        def initialize_log(self, logfile=None, data_dir=None, process_id=-1):
             return
         def get_time(self):
             return
         def enter_event(self):
             return
         def exit_event(self):
             return
-        def log_event(self,  **kwargs):
+        def log_event(self, name, cat, start_time, duration, string_args=None):
```
hariharan-devarajan, line 71: Revert this

hariharan-devarajan (changes requested):
We are almost there. I see we have some code which got reverted from the mock logging.

zhenghh04:
@hariharan-devarajan , I am hesitating between using workflow.log_level to control the log level, vs using DLIO_LOG_LEVEL environment variable to control it. I slightly lean towards to the latter one, which is more common in other apps. What is your preference?

hariharan-devarajan:
> @hariharan-devarajan , I am hesitating between using workflow.log_level to control the log level, vs using DLIO_LOG_LEVEL environment variable to control it. I slightly lean towards to the latter one, which is more common in other apps. What is your preference?

I see a value in both. But I prefer the environment variable, too. By default, we should have WARN level, not info. Then we can make DLIO_LOG_LEVEL to higher logging levels like INFO and DEBUG.

We should switch the per epoch time to print and include the sample rate per epoch there.

Then, per step becomes info. 

And variable logging becomes debug.

zhenghh04:
@hariharan-devarajan After reading the documentation more: https://docs.python.org/3/library/logging.html#levels, I feel that we should have default logging level to be info, and everything should go through logging, not through print. 

What do you think?

hariharan-devarajan:
So printing is different than logging in my opinion. 

The things that tell us about benchmark high-level like initialize, progress, and metrics. 

Logging here is for internal parts. 

I wrote a logger in the past on c++ in which I added print as the highest log above error. 

For benchmarks I feel it makes sense to have this.

zhenghh04:
@hariharan-devarajan How about we just keep the logging system as this PR? We can have another PR to improve it? 

I need this feature for now to have an easy way to reduce some I/O overhead from the logging.

dlio_benchmark/utils/utility.py
zhenghh04: Forgot to response this: 

I tried (**kwargs) does not work. So I revert it back.

zhenghh04:
> We should add error logging as an option for pure performance mode. For the final per metric we should use a different logging, which outputs PRINT messages to STDOUT and redirects to the dlio.log file.
> 
> The other logging is optional from the benchmark.

Now it is doing that

zhenghh04:
> @zhenghh04 Can we not print the file name and line number for info logging. We should only do that for debug logging. This will significantly reduce log size.

This is done!

zhenghh04:
> Now if u want to use logging for printing, I would create two loggers. One for printing and one for logging internal stuff.
> 
> The printing could always be info, internal, and cannot change by benchmark parameters.

Done!

hariharan-devarajan (changes requested):
The documentation still refers to debug in workflow.

The rest looks good. 

https://github.com/search?q=repo%3Aargonne-lcf%2Fdlio_benchmark+debug+path%3A*.rst&type=code

zhenghh04:
> The documentation still refers to debug in workflow.
> 
> The rest looks good.
> 
> https://github.com/search?q=repo%3Aargonne-lcf%2Fdlio_benchmark+debug+path%3A*.rst&type=code

Just fixed this.

hariharan-devarajan (approved):
Looks good.

PR #230, opened 2024-10-05
New improved modelling for LLM Deepspeed.

hariharan-devarajan:
The logic is as follows now. 

Assume we have 40 layers with tensor parallelism of 4 and pipeline parallelism of 8
Then, the checkpointing would have 44 layers (40 + 4 tensor pipeline layers) spread across every 32 ranks.
So, given pipeline_rank being every four ranks in this case, rank 0-3 is pipeline rank 0, 4-7 is pipeline rank 1, and so on.
 
Then, I expect a layer distribution among each pipeline rank to be
(pipeline_rank, start_layer_index, end_layer_index) both the start and end are inclusive.
(0, 0, 5)
(1, 6, 11)
(2, 12, 17)
(3, 18, 23)
(4, 24, 28)
(5, 29, 33)
(6, 34, 38)
(7, 39, 43)
 
Also, a tensor parallelism of 4 would mean each layer tensor would be divided by four on each rank.
So if a layer was (1MB, 1GB) tensors
They would be stored as 256KB, 256MB tensors by each rank.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,20 @@ jobs:
           source ${VENV_PATH}/bin/activate
           pip install --upgrade pip
           pip install -r requirements.txt
+      - name: test_checkpoint_epoch
+        run: |
+          source ${VENV_PATH}/bin/activate
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers0-2-layer_params0-all_ranks] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers1-2-layer_params1-all_ranks] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers2-2-layer_params2-rank_zero] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers3-2-layer_params3-rank_zero] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-all_ranks] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-all_ranks] -v
+          rm -rf data
+      - name: test_checkpoint_step
+        run: |
+          source ${VENV_PATH}/bin/activate
+          mpirun -np 2 pytest -k test_checkpoint_step -v
       - name: test_gen_data
         run: |
           source ${VENV_PATH}/bin/activate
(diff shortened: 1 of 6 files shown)


wvaske:
I hit one issue while testing this. If the checkpoint files did not exist, I would see writes after doing the checkpoint and a comm.barrier(). If the checkpoint files DID exist and I was overwriting them I didn't see this behavior.

I was abled to "fix" this by adding a fsync in the pytorch_checkpointing.py file, but I'm not sure if that's the best way to fix this or if it's a system issue but it ensures that the checkpoint is written as a blocking operation.

```
    @dlp.log
    def save_state(self, suffix, state):
        name = self.get_name(suffix)
        with open(name, "wb") as f:
            torch.save(state, f)
            os.fsync(f.fileno())
            f.close()
```

hariharan-devarajan:
Which file system your on? I tested this on lustre and it was working fine. Maybe the file system synchronization is different on your file system.

From: Wes Vaske ***@***.***>
Date: Monday, October 7, 2024 at 3:32 PM
To: argonne-lcf/dlio_benchmark ***@***.***>
Cc: Hariharan Devarajan ***@***.***>, Author ***@***.***>
Subject: Re: [argonne-lcf/dlio_benchmark] New improved modelling for LLM Deepspeed. (PR #230)

I hit one issue while testing this. If the checkpoint files did not exist, I would see writes after doing the checkpoint and a comm.barrier(). If the checkpoint files DID exist and I was overwriting them I didn't see this behavior.

I was abled to "fix" this by adding a fsync in the pytorch_checkpointing.py file, but I'm not sure if that's the best way to fix this or if it's a system issue but it ensures that the checkpoint is written as a blocking operation.

    @dlp.log

    def save_state(self, suffix, state):

        name = self.get_name(suffix)

        with open(name, "wb") as f:

            torch.save(state, f)

            os.fsync(f.fileno())

            f.close()

hariharan-devarajan:
> I hit one issue while testing this. If the checkpoint files did not exist, I would see writes after doing the checkpoint and a comm.barrier(). If the checkpoint files DID exist and I was overwriting them I didn't see this behavior.
> 
> I was abled to "fix" this by adding a fsync in the pytorch_checkpointing.py file, but I'm not sure if that's the best way to fix this or if it's a system issue but it ensures that the checkpoint is written as a blocking operation.
> 
> ```
>     @dlp.log
>     def save_state(self, suffix, state):
>         name = self.get_name(suffix)
>         with open(name, "wb") as f:
>             torch.save(state, f)
>             os.fsync(f.fileno())
>             f.close()
> ```

I am hesitating for doing sync as it will significantly slow down the system. Can u describe the filesystem on which your writing the checkpoints?

We probably need a flag in dlio_benchmark to be enable fsync for some filesystems.

wvaske:
> > I hit one issue while testing this. If the checkpoint files did not exist, I would see writes after doing the checkpoint and a comm.barrier(). If the checkpoint files DID exist and I was overwriting them I didn't see this behavior.
> > I was abled to "fix" this by adding a fsync in the pytorch_checkpointing.py file, but I'm not sure if that's the best way to fix this or if it's a system issue but it ensures that the checkpoint is written as a blocking operation.
> > ```
> >     @dlp.log
> >     def save_state(self, suffix, state):
> >         name = self.get_name(suffix)
> >         with open(name, "wb") as f:
> >             torch.save(state, f)
> >             os.fsync(f.fileno())
> >             f.close()
> > ```
> 
> I am hesitating for doing sync as it will significantly slow down the system. Can u describe the filesystem on which your writing the checkpoints?
> 
> We probably need a flag in dlio_benchmark to be enable fsync for some filesystems.

I'm using XFS with a single local NVMe drive. I'm OK tracking this change in my local branch for now until I can better confirm if it's a real issue or an artifact of some system configuration issue.

zhenghh04 (approved):

PR #231, opened 2024-10-15
fix misleading generator message

rayandrew:

The current message says `Generating NPZ data` when it should say `Generating HDF5 data`. 
This can be misleading for users.


hariharan-devarajan (approved):
Thanks for the catch

Issue #232, opened 2024-10-16
Negative computation time when stdev of computation time is provided


rayandrew:
Hi @zhenghh04 and @hariharan-devarajan,

I encountered error when providing stdev in the configuration.

We can fix it by adding the `abs` while calling `random.normal` in `main.py`

```diff
[INFO] Computation time is: 1.2056848216668536 [/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py:271]
+ [INFO] Computation time is: -0.22297661682373804 [/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py:271]
[INFO] Computation time is: 0.7074011037224097 [/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py:271]
[INFO] Computation time is: 0.5691646364965793 [/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py:271]
Error executing job with overrides: ['workload=stormer_a100']
Traceback (most recent call last):
  File "/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py", line 387, in run_benchmark
    benchmark.run()
  File "/lus/eagle/projects/MDClimSim/rayandrew/venvs/dlio-benchmark/lib/python3.11/site-packages/dftracer/logger.py", line 214, in wrapper
    x = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py", line 339, in run
    steps = self._train(epoch)
            ^^^^^^^^^^^^^^^^^^
  File "/lus/eagle/projects/MDClimSim/rayandrew/venvs/dlio-benchmark/lib/python3.11/site-packages/dftracer/logger.py", line 214, in wrapper
    x = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/lus/eagle/projects/MDClimSim/rayandrew/dlio-benchmark/dlio_benchmark/main.py", line 272, in _train
    self.framework.compute(batch, epoch, block_step, computation_time)
[...]

PR #233, opened 2024-10-16
fix negative value of computation time when stdev exists

rayandrew:
Due to the nature of normal random equation, the value generated can be negative.

Enforcing it to be positive using `abs` will fix the issue.

Fix: #232


hariharan-devarajan (approved):
Looks good

zhenghh04 (approved):

Issue #235, opened 2024-10-16
Fix last step is not executed with/without user specifying `total_training_steps`


rayandrew:
Hi @zhenghh04 and @hariharan-devarajan,

The bug description is as follow:
- Assume we have 730 steps, the last step is not executed. Thus, only got executed until 729.
- Same problem persists when we specify `total_training_steps` in the configuration

PR #236, opened 2024-10-16
fix last step is not executed

rayandrew:
If we have 730 steps, DLIO benchmark only executes until 729

The bug also persists when user specified `total_training_steps`

Fix: #235


hariharan-devarajan (approved):
Looks good

rayandrew:
There is one bug. if `total_training_steps` is not specified, the default will be -1.
I added  check for that as well

hariharan-devarajan (changes requested):
The CI is failing. Can u run pytest and check.

rayandrew:
I think the last commit fixed the CI @hariharan-devarajan

hariharan-devarajan (approved):
The changes look good. 
1. We make sure we complete the step correctlyly.
2. For DALI, we need to handle the stop iteration gracefully.

hariharan-devarajan:
@zhenghh04 This is ready for merge as well.

zhenghh04 (approved):

PR #238, opened 2024-10-23
fix wrong tracing location of fetch data

rayandrew:

If we use `dlp.iter` inside `main.py`, this does not reflect the real fetch data since the data is already fetched inside `DataLoader.next()`. Because of this, we ended up tracing the fetching of batch data that resides in the memory instead of tracing its fetching duration from filesystem or other sources.

This PR addresses the problem where we put `dlp.iter` inside the dataloader, as close as possible to the dataset itself.

For dali dataloader, I haven't implemented the fix since there is no visible iteration there (DALI uses `share_outputs`)

Note: @hariharan-devarajan when we bump the dftracer version, we should provide the name of the `dlp.iter`.
I left a note to change `dlp.iter` to `dlp.iter(..., name=self.next.__qualname__)`

### DIFF
--- a/dlio_benchmark/data_loader/native_dali_data_loader.py
+++ b/dlio_benchmark/data_loader/native_dali_data_loader.py
@@ -60,7 +60,9 @@ class NativeDaliDataLoader(BaseDataLoader):
             pipeline.reset()
         for step in range(num_samples // batch_size):
             try:
-                for batch in self._dataset:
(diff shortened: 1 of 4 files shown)


hariharan-devarajan (approved):
This looks good thanks

zhenghh04 (approved):

PR #240, opened 2024-10-24
Change `max` to `abs` for preprocess time

rayandrew:

This PR addresses the problem where `random.normal` result can be negative.
if we do `max(t, 0)` then we will reduce the variance of preprocess time when the generated value is negative. 

It is better to use `abs` in this case.

At the same time, I add strict comparison to make sure user does not give negative value while specifying `preprocess_time` or `preprocess_time_stdev` in the config.


hariharan-devarajan (approved):
This looks good.

PR #242, opened 2024-11-12
upgrade pydftracer package

rayandrew:

This PR will update pydftracer package from `1.0.2` to `1.0.6`.

Note that pydftracer package `1.0.6` still does not include the new API of `dlp.iter(self, func, name, iter_name)` that is needed for [torch data loader](https://github.com/argonne-lcf/dlio_benchmark/blob/main/dlio_benchmark/data_loader/torch_data_loader.py#L169) but it is not a big of a deal for now.

Thanks!

### DIFF
--- a/dlio_benchmark/main.py
+++ b/dlio_benchmark/main.py
@@ -70,7 +70,7 @@ class DLIOBenchmark(object):
             <li> local variables </li>
         </ul>
         """
-        global dftracer, dftracer_init, dftracer_finalize
+        global dftracer, dftracer_initialize, dftracer_finalize
 
         t0 = time()
         self.args = ConfigArguments.get_instance()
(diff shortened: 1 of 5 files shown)


zhenghh04:
@rayandrew Could you fix the github action issue?

rayandrew:
Hi @zhenghh04,

@hariharan-devarajan and I have fixed the double corruption issue.

The problem in test_subset was caused by run_benchmark being called twice:
- The first call initializes and finalizes dftracer.
- The second call re-initializes dftracer again.

When dlio_benchmark tries to clean up resources during the second call, it attempts to free memory that has already been finalized, leading to corruption.

The fix is simple: we just need to have global variable to denote whether we want to initialize/finalize dftracer or not

rayandrew:
ok we have another problem with `test_computation_time_distribution`, I will try to fix that

rayandrew:
done this is ready to review @zhenghh04 and @hariharan-devarajan

zhenghh04 (approved):
This looks good to me. 

@hariharan-devarajan, Any further comment? If not, I will merge it.

PR #250, opened 2025-02-06
Support O_DIRECT dataloader for pytorch 

YardenMa:
Adding support to use O_Direct data-loader by specifying reader.data_loader = pytorch-direct in the workload configuration file

### DIFF
--- a/dlio_benchmark/common/enumerations.py
+++ b/dlio_benchmark/common/enumerations.py
@@ -158,6 +158,7 @@ class DataLoaderType(Enum):
     """
     TENSORFLOW='tensorflow'
     PYTORCH='pytorch'
+    PYTORCH_DIRECT='pytorch-direct'
     DALI='dali'
     NATIVE_DALI='native_dali'
     CUSTOM='custom'
--- a/dlio_benchmark/data_loader/data_loader_factory.py
+++ b/dlio_benchmark/data_loader/data_loader_factory.py
@@ -37,7 +37,7 @@ class DataLoaderFactory(object):
             if DLIOMPI.get_instance().rank() == 0:
                 logging.info(f"{utcnow()} Running DLIO with custom data loader class {_args.data_loader_class.__name__}")
             return _args.data_loader_class(format_type, dataset_type, epoch)
-        elif type == DataLoaderType.PYTORCH:
+        elif type == DataLoaderType.PYTORCH or type == DataLoaderType.PYTORCH_DIRECT:
             from dlio_benchmark.data_loader.torch_data_loader import TorchDataLoader
             return TorchDataLoader(format_type, dataset_type, epoch)
         elif type == DataLoaderType.TENSORFLOW:
--- a/dlio_benchmark/reader/odirect_reader.py
+++ b/dlio_benchmark/reader/odirect_reader.py
@@ -1,0 +1,180 @@
+"""
+   Copyright (c) 2024, UChicago Argonne, LLC
+   All Rights Reserved
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+"""
+import numpy as np
+
+import os 
+import ctypes
+import time
+import struct
+import zlib
+
+from dlio_benchmark.common.constants import MODULE_DATA_READER
+from dlio_benchmark.reader.reader_handler import FormatReader
+from dlio_benchmark.utils.utility import Profile
+
+dlp = Profile(MODULE_DATA_READER)
+
+
+class ODirectReader(FormatReader):
+    """
+    ODirectReader - Generic reader for files using O_DIRECT
+    """
+
+    @dlp.log_init
+    def __init__(self, dataset_type, thread_index, epoch, parser=None, alignment=4096):
+        super().__init__(dataset_type, thread_index)
+        self.parser = parser
+        self.alignment = alignment
+
+    @dlp.log
+    def open(self, filename):
+        super().open(filename)
+        data =  self.odirect_read(filename)
+        if self.parser:
+            data = self.parser(data)
+        return data
+
+    def odirect_read(self, filepath):
+        try:
+            # Open the file with O_DIRECT
+            fd = os.open(filepath, os.O_RDONLY | os.O_DIRECT)
+
+            # Get the file size
+            file_size = os.path.getsize(filepath)
+
+            # Calculate the buffer size, aligned to the given alignment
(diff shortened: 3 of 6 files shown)


zhenghh04 (commented):
Could you add a github action to test this feature?

hariharan-devarajan:
@YardenMa I have a fundamental question, why is it called PYTORCH O_DIRECT? the implementation is simple a O_DIRECT for reading npy files. 

We should simply add it as a new format, add reader and that should be it. We dont need PYTORCH_DIRECT for this.

dlio_benchmark/common/enumerations.py
```diff
@@ -158,6 +158,7 @@ class DataLoaderType(Enum):
     """
     TENSORFLOW='tensorflow'
     PYTORCH='pytorch'
+    PYTORCH_DIRECT='pytorch-direct'
```
hariharan-devarajan, line 161: This is not needed switch this to a new format like NPY_O_DIRECT.

dlio_benchmark/data_loader/data_loader_factory.py
```diff
@@ -37,7 +37,7 @@ def get_loader(type, format_type, dataset_type, epoch):
             if DLIOMPI.get_instance().rank() == 0:
                 logging.info(f"{utcnow()} Running DLIO with custom data loader class {_args.data_loader_class.__name__}")
             return _args.data_loader_class(format_type, dataset_type, epoch)
-        elif type == DataLoaderType.PYTORCH:
+        elif type == DataLoaderType.PYTORCH or type == DataLoaderType.PYTORCH_DIRECT:
```
hariharan-devarajan, line 40: Then this check would be for checking if type if NPY_O_DIRECT as the reader.

dlio_benchmark/reader/odirect_reader.py
```diff
@@ -0,0 +1,180 @@
...
+import zlib
+
+from dlio_benchmark.common.constants import MODULE_DATA_READER
+from dlio_benchmark.reader.reader_handler import FormatReader
+from dlio_benchmark.utils.utility import Profile
+
+dlp = Profile(MODULE_DATA_READER)
+
+
+class ODirectReader(FormatReader):
```
hariharan-devarajan, line 32: Is this generic or is specific for npy files? Your enumeration choice indicates it's specific.

dlio_benchmark/reader/reader_factory.py
```diff
@@ -55,15 +55,24 @@ def get_reader(type, dataset_type, thread_index, epoch_number):
...
                 from dlio_benchmark.reader.npy_reader import NPYReader
-                return NPYReader(dataset_type, thread_index, epoch_number)                         
+                return NPYReader(dataset_type, thread_index, epoch_number)                 
         elif type == FormatType.NPZ:
             if _args.data_loader == DataLoaderType.NATIVE_DALI:
                 raise Exception("Loading data of %s format is not supported without framework data loader; please use npy format instead." %type)
+            elif _args.data_loader == DataLoaderType.PYTORCH_DIRECT:
+                from dlio_benchmark.reader.odirect_reader import ODirectReader, parse_npz
+                def parse_npz_x(mem_view):
+                    return parse_npz(mem_view)["x"]
```
hariharan-devarajan, line 70: We should not have any reader specific code here. I would move this to init function itself.

dlio_benchmark/utils/config.py
```diff
@@ -272,7 +272,7 @@ def derive_configurations(self, file_list_train=None, file_list_eval=None):
         if self.data_loader_sampler is None and self.data_loader_classname is None:
             if self.data_loader == DataLoaderType.TENSORFLOW:
                 self.data_loader_sampler = DataLoaderSampler.ITERATIVE
-            elif self.data_loader in [DataLoaderType.PYTORCH, DataLoaderType.DALI]:
+            elif self.data_loader in [DataLoaderType.PYTORCH, DataLoaderType.PYTORCH_DIRECT, DataLoaderType.DALI]:
```
hariharan-devarajan, line 275: Again, this shows it has nothing to do with PYTORCH. Lets fix the overall design here.

hariharan-devarajan (changes requested):
I think the code is specific to NPY accessed through O_DIRECT. I recommend to do a new format which specifies the NPY and O_DIRECT and then do the implementation. I am not sure if it needs to be specific to PYTORCH.

dlio_benchmark/common/enumerations.py
YardenMa: Please see my comment in the main thread and let me know if you agree or not.

dlio_benchmark/data_loader/data_loader_factory.py
YardenMa: I did not see a way to specify the reader from the configuration file, is that possible ?

dlio_benchmark/reader/odirect_reader.py
YardenMa: The reader is generic, the parsing function in the constructor is specific

dlio_benchmark/reader/reader_factory.py
YardenMa: This is here because this logic is specific to the way npz files are generated. Each .npz file that's generate (for 3dunet at least) has 2 files in it "x" and "y". The NPZReader is running np.load(file)["x"]. 
As the ODirectReader and prase_npz functions are generic, I wrapped the specific reference to "x" here. Where do you think I should put that assuming ODirectReader is generic ?

YardenMa:
@zhenghh04 @hariharan-devarajan 
Please let me know how you want me to move forward with this one given my previous answer

zhenghh04:
@YardenMa How about we do the following which will make the code easier to maintain in future? 

* In the yaml file, we allow people to set o_direct to True, if they would like to use that. 
```yaml
reader: 
 ..
 odirect: True
```
* In the implementation, please create another npz_reader called npz_reader_odirect. in the reader_handler, set the reader to be ```npz_reader_odirect``` if ```reader.odirect==True``` and ```reader.format==npz```. 
* Please also add exception message if odirect==True, but format != npz || npy. 
* In this way, we do not have to use PYTORCH_ODIRECT loader at all.

zhenghh04 (changes requested):
@YardenMa How about we do the following which will make the code easier to maintain in future? 

* In the yaml file, we allow people to set o_direct to True, if they would like to use that. 
```yaml
reader: 
 ..
 odirect: True
```
* In the implementation, please create another npz_reader called npz_reader_odirect. in the reader_handler, set the reader to be ```npz_reader_odirect``` if ```reader.odirect==True``` and ```reader.format==npz```. 
* Please also add exception message if odirect==True, but format != npz || npy. 
* In this way, we do not have to use PYTORCH_ODIRECT loader at all.

timohty-chau:
Wanted to follow up and see if this change will be merged into the main branch soon as I've been able to test in our labs and the benefit of running more accelerator per client would help reduce the number of servers needed for a given accelerator count.

zhenghh04:
@YardenMa , do you have any cycles to work on this? We have to get this in by next Friday, May 2, 2025, in order to make it available for MLPerf Storage v2.0 submission.

zhenghh04:
@YardenMa Thanks for your effort here. @timohty-chau added the changes. Now the odirect feature is good to be merged.

zhenghh04:
I'll close this PR as all of this is contained in #286

PR #251, opened 2025-02-07
fixed bug on doc action

zhenghh04:
This is to fixed the issue for recent requirement on deploying documentation. Without this fix, the github actions are failing.


zhenghh04:
@hariharan-devarajan This is a small fix. I'll just merge it directly. Let me know if you have problem with it. We can always revert.

PR #259, opened 2025-02-25
left over logging fix

zhenghh04:
This is just to fix some left over logging.debug, logging.info after merging. 

I also removed the doc deployment which is causing failure. We already make the deployment through readthedoc.


zhenghh04:
@hariharan-devarajan Hi Hari, could you take a look of this PR. Should be simple.

hariharan-devarajan (approved):
Looks good

Issue #244, opened 2024-12-08
Docker images on dockerhub are seriously outdated.


russfellows:
The docker image referenced in the Readme is seriously out of date, it is pre- v0.5, and was pushed 2 years ago.  Also, the image with the tag "latest" was built for ARM64, which is likely not the architecture most people would be expecting.  

The Readme, Docker building process and images need some work in order to make them relevant again, circa 2025.

glimchb:
see https://github.com/argonne-lcf/dlio_benchmark/pull/263

glimchb:
@zhenghh04 please close as fixed

PR #263, opened 2025-02-27
ci: add docker build and publish

glimchb:
addresses #244
docker fix in #265

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,56 @@
+---
+name: Docker
+
+on:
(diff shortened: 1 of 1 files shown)


zhenghh04:
@glimchb , Looks like this is still failing because of DALI data loader module. Will take a look and see whether we can get around it.

glimchb:
> @glimchb , Looks like this is still failing because of DALI data loader module. Will take a look and see whether we can get around it.

I fixed it in #265 just need to rebase and it will pass...

PR #264, opened 2025-02-28
S3 support with Boto3 for Pytorch NPZ data a generator and dataloader

ekaynar:
PytorchS3Storage class, which leverages the Boto3 library to interact with S3 storage. Adding support to data generation and loading of NPZ files.

### DIFF
--- a/dlio_benchmark/data_generator/npz_generator.py
+++ b/dlio_benchmark/data_generator/npz_generator.py
@@ -1,4 +1,5 @@
 """
+   Copyright (c) 2025 Dell Inc, or its subsidiaries.
    Copyright (c) 2024, UChicago Argonne, LLC
    All Rights Reserved
 
--- a/dlio_benchmark/data_generator/npz_generator.py
+++ b/dlio_benchmark/data_generator/npz_generator.py
@@ -20,11 +20,13 @@ from dlio_benchmark.data_generator.data_generator import DataGenerator
 
 import logging
 import numpy as np
+import io
 
 from dlio_benchmark.utils.utility import progress, utcnow
 from dlio_benchmark.utils.utility import Profile
 from shutil import copyfile
 from dlio_benchmark.common.constants import MODULE_DATA_GENERATOR
+from dlio_benchmark.common.enumerations import StorageType
 
 dlp = Profile(MODULE_DATA_GENERATOR)
 
@@ -51,8 +53,19 @@ class NPZGenerator(DataGenerator):
             out_path_spec = self.storage.get_uri(self._file_list[i])
             progress(i+1, self.total_files_to_generate, "Generating NPZ Data")
             prev_out_spec = out_path_spec
-            if self.compression != Compression.ZIP:
-                np.savez(out_path_spec, x=records, y=record_labels)
+
+            if self._args.storage_type == StorageType.S3:
+                buffer =  io.BytesIO()
+                if self.compression != Compression.ZIP:
+                    np.savez(buffer, x=records, y=record_labels)
+                else:
+                    np.savez_compressed(buffer, x=records, y=record_labels)
+                self.storage.put_data(out_path_spec, buffer)
+
+
             else:
-                np.savez_compressed(out_path_spec, x=records, y=record_labels)
+                if self.compression != Compression.ZIP:
+                    np.savez(out_path_spec, x=records, y=record_labels)
+                else:
+                    np.savez_compressed(out_path_spec, x=records, y=record_labels)
         np.random.seed()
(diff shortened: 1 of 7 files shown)


zhenghh04:
@ekaynar could you try to fix the failing ci tests?

dlio_benchmark/data_generator/npz_generator.py
```diff
@@ -1,4 +1,5 @@
 """
+   Copyright (c) 2025 Dell Inc, or its subsidiaries.
```
hariharan-devarajan, line 2: We cannot have companies copyright here.

hariharan-devarajan:
@ekaynar Thank you for your updates for S3 support. I think its a great addition to DLIO Benchmark

Based on the code guidelines currently followed on DLIO Benchmark, I suggest the following updates.
- [ ] We do not want switch cases in the NPZ Reader. Instead, inherit the NPZ Reader class (NPZS3Reader in a new module called npz_s3_reader) and update the open logic. The reasoning behind this is that it would easier to not install s3 if it's not supported on a specific system. 
- [ ] The `self.storage` variable is defined within NPZS3Reader's `init`.
- [ ] No additional company copyrights as open Apache 2.0 License.
- [ ] Add a new enumeration on StorageType as S3 and do a switch case within ReaderFactory class. 
- [ ] Similar to Reader, inherit the NPZGenerator class for S3 and override the `generate` function.
- [ ] We need to include S3 emulation for the CI to test the above functionality and include test-case for that environment. if your not aware of S3 emulators that can run locally, @zhenghh04 could recommend you some. 

Let me know if this makes sense or if you have concern.

dlio_benchmark/reader/npz_reader.py
```diff
@@ -1,4 +1,5 @@
 """
+   Copyright (c) 2025 Dell Inc, or its subsidiaries.
```
hariharan-devarajan, line 2: We cannot have companies copyright here.

dlio_benchmark/data_generator/npz_generator.py
```diff
@@ -51,8 +54,19 @@ def generate(self):
             out_path_spec = self.storage.get_uri(self._file_list[i])
             progress(i+1, self.total_files_to_generate, "Generating NPZ Data")
             prev_out_spec = out_path_spec
-            if self.compression != Compression.ZIP:
-                np.savez(out_path_spec, x=records, y=record_labels)
+
```
hariharan-devarajan, line 57: Create a new class which inherits NPZGenerator and switch on generator_factory based on storage type.

dlio_benchmark/reader/npz_reader.py
```diff
@@ -34,6 +37,12 @@ def __init__(self, dataset_type, thread_index, epoch):
 
     @dlp.log
     def open(self, filename):
+        if isinstance(self.storage, S3PytorchStorage):
```
hariharan-devarajan, line 40: Move this to a new class that inherits NPZReader and overrides open function.

dlio_benchmark/reader/reader_handler.py
```diff
@@ -44,7 +45,8 @@ def __init__(self, dataset_type, thread_index):
             f"Loading {self.__class__.__qualname__} reader on thread {self.thread_index} from rank {self._args.my_rank}")
         self.dataset_type = dataset_type
         self.open_file_map = {}
-
+        self.storage = StorageFactory().get_storage(self._args.storage_type, self._args.storage_root,
```
hariharan-devarajan, line 48: Move this to new reader class' init function

dlio_benchmark/storage/s3_storage.py
```diff
@@ -1,4 +1,5 @@
 """
+   Copyright (c) 2025 Dell Inc, or its subsidiaries.
```
hariharan-devarajan, line 2: Remove company copyright.

hariharan-devarajan (changes requested):

ekaynar:
> @ekaynar could you try to fix the failing ci tests?

dlio_benchmark/reader/npz_reader.py
ekaynar: @hariharan-devarajan  Thank you for the feedback! I appreciate your suggestions. I will move everything into the NPZS3readers class. 

My suggestion is that, in the long term, having a single reader class for different file types, which then directs to the appropriate storage class (file or object), might be a more sustainable approach compared to having separate readers for different file accesses. This way, we avoid duplicating code for different file types, and a unified structure can make the codebase easier to understand for other developers.

dlio_benchmark/data_generator/npz_generator.py
ekaynar: @hariharan-devarajan, currently, GeneratorFactory only receives the format type (NPZ). Should I pass the storage type during the initialization, so that it can select either NPZS3Generator or NPZGenerator class?

PR #268, opened 2025-03-11
docs: small readme typo

glimchb:
- missing pull command for docker
- use python console scripts like in other examples instead of invoking python on script directly


glimchb:
@zhenghh04 please review

PR #273, opened 2025-03-21
Fixed loading checkpoint timer

zhenghh04:
The start checkpoint timer was not corrected set. It was set to checkpoint save state. This is included in #274


hariharan-devarajan (approved):
Looks good.

PR #278, opened 2025-04-03
Separate read and write checkpoints. 

zhenghh04:
- [x] We do write and read separately instead of doing read in between the writes
- [x] Setting default value of rank shift to be False because people can now avoid caching effect by using larger number of writes and then read them back. 
- [x] Setting fsync to be True by default to measure the pushing data to persistent storage system.


zhenghh04:
@hariharan-devarajan please review this again. After the mlperf storage working group meeting on the LLM checkpointing. I did some new development based on the request.

hariharan-devarajan (approved):
Looks good.

PR #281, opened 2025-04-15
Reduce RAM Usage for checkpointing (zero1)

LouisDDN:
With zero1, only the first DP saves the model parameters. There is no need to allocate these parameters on other DP.

The original RAM formula for zero1 was:
`ram = ((model_size/(TP*PP))+(optimizer_size/(TP*PP*DP)))*(TP*PP*DP)`
Which would scale badly with increased DP.

With the patch, this simplifies to:
`ram = model_size + optimizer_size`
Which is equal to zero3 RAM usage.

The reduction becomes more significant as DP increases. 
For example, a 70B model with 8TP, 4PP, and 4DP requires 1300GB without the patch, but only 910GB with it.

This is still not equivalent to zero3, since only the first DP saves the model parameters.
The RAM usage changed but the IO workload should be the same. 

Measured on a single node 70B: 8 (TP) x 4 (PP) x 4 (DP) = 128
Without the patch: 1.4TiB RAM usage (without accouting for pagecache)
With the patch: 1TiB usage

### DIFF
--- a/dlio_benchmark/checkpointing/base_checkpointing.py
+++ b/dlio_benchmark/checkpointing/base_checkpointing.py
@@ -99,13 +99,20 @@ class BaseCheckpointing(ABC):
                 for layer_index in range(start_layer, end_layer + 1):
                     self.layer_state[str(layer_index)] = layer_state  
             elif self.args.num_layers > 0:
-                self.layer_state = dict()
-                model_checkpoint_size = 0.0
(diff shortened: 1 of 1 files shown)


LouisDDN:
here is the slide explaining what is done:
PR281

LouisDDN:
Present in #283

PR #282, opened 2025-04-16
Add hash consing optimization (checkpoint)

LouisDDN:
Creates two LUTs to decrease RAM usage by reusing identical tensors.
Detailed algorithm explaination is in the save_state function. 
With this patch, we can use an amount of RAM far below the actual model size.
It seems to work. 

**RAM usage with/without + performance comparison**
I removed the Y right axis details, as it represent real performance number in GB/s for read and write, but it start at 0.
The objective here is to demonstrate that the performance is the same, while the RAM usage decreases a lot.
C= chunksize. This value makes sense if you read the detailed explaination. I have set the default at 1. 

hash_consing

**Write verification**
This is the amount of data available on disk after each write test (1 iter)
|         in %            | Normal | hash + C=8 | hash + C=4 | hash + C=2 | hash + C=1 | hash+ C=1 + #281  | Normal + #281 |
|---------------------|--------|------------|------------|------------|------------|--------------|----------------|
| Data on disk after 1 write iter | 100      | 99,99629   | 99,99989   | 99,99592   | 99,99874   | 100,0001     | 99,99937       |

**Raw grafana view**
memory usage (1 write then 1 read, clean cache in between, visible on this graph)
normal, hash+C1, hash+C2,hash+C4, hash+C8
mem

network usage measured on the client (read)
There are two colors but it's just because there are two network interface.
normal, hash+C1, hash+C2,hash+C4, hash+C8
READ_IB

network usage measured on the client (write)
Same
normal, hash+C1, hash+C2,hash+C4, hash+C8
WRITE_IB

**Important**
To make it work, the optimizer state write/read operation are chunked, which is fine from my perspective but it does change the algorithm a bit. It creates multiple files, as it is done with the layers of the model  when DP is above 0. 

Can be enabled/disabled with:
`++workload.checkpoint.hash_consing=True/False (default: False)`

Can be tuned with:
`++workload.checkpoint.hash_consing_chunk_size=1 (default: 1) Increasing the value increases memory usage.
`
Default is the minimum RAM usage.

### DIFF
--- a/dlio_benchmark/checkpointing/base_checkpointing.py
+++ b/dlio_benchmark/checkpointing/base_checkpointing.py
@@ -18,6 +18,7 @@ import logging
 import math
 import os
 import time
+from collections import Counter
 from abc import ABC, abstractmethod
 
 from dlio_benchmark.common.enumerations import CheckpointLocationType, CheckpointModeType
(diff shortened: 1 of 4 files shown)


LouisDDN:
Here is a slide explaining the algorithm
HASHCONSING

PR #283, opened 2025-04-17
RAM optimisations for checkpointing

LouisDDN:
This PR merges both #281 and #282 into a single one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,11 @@ jobs:
           mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-0] -v
           mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-0] -v
           rm -rf data
+      - name: test_checkpoint_ksm_config
+        run: |
+          source ${VENV_PATH}/bin/activate
+          mpirun -np 2 pytest -k test_checkpoint_ksm_config -v
+	  rm -rf data
       - name: test_checkpoint_step
         run: |
           source ${VENV_PATH}/bin/activate
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
         run: |
           source ${VENV_PATH}/bin/activate
           mpirun -np 2 pytest -k test_checkpoint_ksm_config -v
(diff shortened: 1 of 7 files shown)


zhenghh04:
> @zhenghh04 I am working on a new patch as an alternative to hash consing, based on Friday’s discussion. It will increase startup time but fully preserve the original I/O pattern (no chunking).
> 
> I have already implemented it for the write operation. It achieves an even greater memory reduction than hash consing. The read operation may prove more complex; the details are not yet clear.

Hi @LouisDDN , are you ready for the review? If you commit the new change, @hariharan-devarajan and I I will take a look.

hariharan-devarajan (changes requested):
@zhenghh04 The changes look good. I think, we should create a `ksm` tree within the checkpointing and then have all its options there. Then we can derive `ksm_init` ourself if `ksm` tree was enabled. Once we work out the configuration changes, please add that to the documents. Once both are done. I can take a final look and approve.


- [ ] Test checking the functionality
- [ ] Improvement to Configuration
- [ ] Documentation

LouisDDN:
RAM usage comparison: `++workload.checkpoint.ksm_init=false` vs. `++workload.checkpoint.ksm_init=true`.
Execution order: 8 B Zero3 (false), 70 B Zero3 (false), 8 B Zero3 (true), 70 B Zero3 (true).

RAM_usage_KSM_8b_70b


Single-node RAM usage with 405 B; this run would not have been possible without this PR.
Left: physical RAM usage, which never exceeds 1 TB. Right: virtual memory usage, which exceeds 6 TB.
Initialization takes several hours for this reduction, but it is feasible, as demonstrated.

RAM_usage_KSM_405B

It divides the memory requirement by 6. 

With the 6x factor, 1T parameter tests can run on 32 clients with 128 GB each. This applies only to checkpoint writes and does not reduce memory requirements for reads.

LouisDDN:
@hariharan-devarajan @zhenghh04 Thank you for your feedback.
I did the fix:
- KSM config nested
- Derived ksm_init
- Docs updated
- Test added

zhenghh04:
@LouisDDN , could you also add tests to .github/workflows/ci.yml?

LouisDDN:
@zhenghh04  done

zhenghh04 (approved):

zhenghh04 (approved):

zhenghh04:
Thanks @LouisDDN for all the effort here!

PR #284, opened 2025-04-25
configs

zhenghh04:
This is minor changes to the config files; to reflect the request from the MLPerf Storage.


hariharan-devarajan (approved):
Looks good.

PR #288, opened 2025-05-06
Fix issues with tensorflow tf generator support for Object Storage

annie-anna:
Found a few bugs when I run MLPerf storage on Object Storage:

1. tfrecord index files are still pointing to local FS path rather than object path. 
    - Update index file paths to include ``s3://``.
2. When creating node, unable to create intermediate directories. 
    - Use ``makedirs`` instead of ``mkdir`` to create intermediate dirs.
3. ``walk_node`` is using ``tf.io.gfile.listdir``, which has a bug that causes out-of-index for substr.
    - Use boto3 ``list_object_v2``.

### DIFF
--- a/dlio_benchmark/framework/tf_framework.py
+++ b/dlio_benchmark/framework/tf_framework.py
@@ -31,7 +31,9 @@ from dlio_benchmark.common.enumerations import FrameworkType, Profiler, FormatTy
     DataLoaderType
 
 import tensorflow as tf
+import tensorflow_io as tfio
 from tensorflow.python.framework import errors
+import boto3
 
 tf.compat.v1.logging.set_verbosity(tf.compat.v1.logging.ERROR)
 
@@ -52,6 +54,9 @@ class TFFramework(Framework):
             else:
                 self.tensorboard = ProfilerFactory.get_profiler(Profiler.TENSORBOARD)
         self.reader_handler = None
+        self.s3 = boto3.client("s3", \
+                               endpoint_url=os.getenv('S3_ENDPOINT'), \
+                               region_name=os.getenv('AWS_REGION'))
 
     @dlp.log
     def init_loader(self, format_type, epoch=0, data_loader=None):
@@ -102,7 +107,7 @@ class TFFramework(Framework):
 
     @dlp.log
     def create_node(self, id, exist_ok=False):
-        tf.io.gfile.mkdir(id)
+        tf.io.gfile.makedirs(id)
         return True
 
     @dlp.log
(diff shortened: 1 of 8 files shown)


zhenghh04:
@johnugeorge , could you take a look and see whether the changes look good to you?

zhenghh04:
@hariharan-devarajan please review this PR also.

dlio_benchmark/framework/tf_framework.py
```diff
@@ -119,7 +124,24 @@ def get_node(self, id):
     def walk_node(self, id, use_pattern=False):
         try:
             if not use_pattern:
-                return tf.io.gfile.listdir(id)
+                # parse id to get bucket name and prefix
```
johnugeorge, line 127: Can you explain the motivation in overriding the function?

johnugeorge: Can you explain the bug that you referred to?

annie-anna: Because when I run ``tf.io.gfile.listdir``, I encountered an IndexError, saying the position passed into ``s.substr()`` is greater than ``s.size()``.
I looked into tensorflow code, ``tf.io.gfile.listdir`` is defined by ``list_directory_v2()``, which calls ``GetChildren()``. ``GetChildren()`` is defined in tensorflow/io and will basically return all the objects in the bucket with the given prefix. This func tries to do ``s.substr(s.size()+1)``. There is a detailed code walk in https://github.com/tensorflow/io/pull/2149 that explains the bug.

zhenghh04: @johnugeorge does this look good to you? Are you able to do a quick test?

setup.py
```diff
@@ -17,11 +17,13 @@
     "omegaconf>=2.2.0",
     "pandas>=1.5.1",
     "psutil>=5.9.8",
+    "boto3",
 ]
 x86_deps = [
     f"hydra-core>={HYDRA_VERSION}",
     "nvidia-dali-cuda120>=1.34.0",
-    "tensorflow>=2.11.0",
+    "tensorflow==2.13.1",
```
zhenghh04, line 25: Could we set tensorflow>=2.13.1?

zhenghh04:
@annie-anna The github action jobs are failing because of the dependency issue. tensorflow==2.13.1 and torch>=2.2.0 have incompatible dependencies.

Could you check whether changing to tensorflow>=2.13.1 (and also releasing the constrain for tensorflow_io) still works for the S3 storage?

annie-anna:
> @annie-anna The github action jobs are failing because of the dependency issue. tensorflow==2.13.1 and torch>=2.2.0 have incompatible dependencies.
> 
> Could you check whether changing to tensorflow>=2.13.1 (and also releasing the constrain for tensorflow_io) still works for the S3 storage?

setup.py
annie-anna: See comment below.

annie-anna:
@zhenghh04 @johnugeorge Is there any other question or concern regarding the PR?

dlio_benchmark/framework/tf_framework.py
johnugeorge: Won't this break the storage/framework abstraction? Right now, this PR will tie S3 storage with the tf framework.

zhenghh04: @johnugeorge yes, but this is just to fix the original implementation, right? The original implementation for tf framework as well. All the create_node, walk_node functions are part of the framework class. 

For pytorch, we can implement the same storage functions under pytorch_framework class.

annie-anna: @zhenghh04 @johnugeorge ack, I reverted changes in ``walk_node``.

zhenghh04:
Thanks very much @annie-anna.

annie-anna:
@zhenghh04 Why did the build fail again?

zhenghh04:
@annie-anna , I incorporated your changes in the most recent PR #294 . Could you try to pull the most recent changes from main?

PR #291, opened 2025-05-15
Randomize tensor data by default (checkpoint)

LouisDDN:
Addresses Russ's discussion on using tensors filled with ones for checkpointing for MLPerf storage.

This PR changes the default behavior to use randomized tensor data for checkpoints.
To revert to tensors filled with ones, use the Hydra override:
`++workload.checkpoint.randomize_tensor=false`

The PR hasn't been tested yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,12 +87,12 @@ jobs:
           mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers3-2-layer_params3-3-True] -v
           mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-0-True] -v
           mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-0-True] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers0-2-layer_params0-0-False] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers1-2-layer_params1-0-False] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers2-2-layer_params2-3-False] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers3-2-layer_params3-3-False] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-0-False] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers6-2-layer_params6-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers7-2-layer_params7-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers8-2-layer_params8-3-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers9-2-layer_params9-3-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers10-1-layer_params10-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers11-1-layer_params11-0-False] -v
           rm -rf data
       - name: test_checkpoint_ksm_config
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,12 +81,18 @@ jobs:
       - name: test_checkpoint_epoch
         run: |
           source ${VENV_PATH}/bin/activate
-          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers0-2-layer_params0-0] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers1-2-layer_params1-0] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers2-2-layer_params2-3] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers3-2-layer_params3-3] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-0] -v
-          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-0] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers0-2-layer_params0-0-True] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers1-2-layer_params1-0-True] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers2-2-layer_params2-3-True] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers3-2-layer_params3-3-True] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-0-True] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-0-True] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers0-2-layer_params0-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers1-2-layer_params1-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers2-2-layer_params2-3-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers3-2-layer_params3-3-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[tensorflow-1024-optimizers4-1-layer_params4-0-False] -v
+          mpirun -np 2 pytest -k test_checkpoint_epoch[pytorch-1024-optimizers5-1-layer_params5-0-False] -v
           rm -rf data
       - name: test_checkpoint_ksm_config
         run: |
--- a/dlio_benchmark/checkpointing/base_checkpointing.py
+++ b/dlio_benchmark/checkpointing/base_checkpointing.py
@@ -71,6 +71,7 @@ class BaseCheckpointing(ABC):
         self.rank_to_checkpoint = self.args.my_rank
         self.num_parameters = self.get_num_parameters()
         self.checkpoint_size = 0.0
+        self.randomize_tensor = self.args.checkpoint_randomize_tensor
 
         # KSM optim
         self.madvise_initialized = False
@@ -192,7 +193,7 @@ class BaseCheckpointing(ABC):
         return False # Default behavior if not overridden
 
     @abstractmethod
-    def get_tensor_core(self, length, datatype="int8"):
+    def get_tensor_core(self, length, datatype="int8", randomize=True):
         return []
 
     def init_madvise(self):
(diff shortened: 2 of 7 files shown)


FileSystemGuy:
Depending upon how the `torch.rand()` and `torch.randint()` functions are implemented, this could be a performance problem.  If they're just a typical pseudo-random number generator, then OK, but if they're fronting a cryptographic-quality-randomness number generator then they will pull entropy info from the kernel, which will quickly deplete, and we'll have to block until more entropy can be acquired.  I like the option to enable random data or not, but IMHO we should check the performance before accepting the change.

zhenghh04:
@LouisDDN I made it into ones instead of random, because it looks like creating random tensors are computationally expensive. 

Also if we do random tensor, will it break your KSM?

LouisDDN:
> @LouisDDN I made it into ones instead of random, because it looks like creating random tensors are computationally expensive.
> 
> Also if we do random tensor, will it break your KSM?

It won't completely break KSM, but it will essentially do nothing instead of reducing RAM usage.

The idea of including it as a toggleable option in this PR is to defer the decision to MLPerf on whether the data needs to be randomized, without requiring subsequent code modifications. See the email thread with Russ for more details.

LouisDDN:
It’s pseudo-random, thankfully.  

I tested this PR on a single client. During testing, the allocation time was identical with or without randomization. The compression ratio is much closer to that of a real PT file (it’s even lower; see graph).  

tensorrand


The seed for the pseudo-random generation is unique per MPI process by default, so no deduplication/compression is possible across processes. 
```verbatim
  Data collected from all 4 processes (at Rank 0 for Part 2 context):
    Process 0: Reported Initial Seed = 474913267114404132, torch.rand() val = 0.85287416
    Process 1: Reported Initial Seed = 7938182670653440165, torch.rand() val = 0.16578156
    Process 2: Reported Initial Seed = 270863154063184961, torch.rand() val = 0.75205654
    Process 3: Reported Initial Seed = 11174190051958333003, torch.rand() val = 0.35592496
```

In terms of performance, using a single client, the difference is minimal (write-only) with or without randomization. Deduplication and compression were disabled on the tested parallel file system.

image

LouisDDN:
I didn't read the correct timing for memory allocation, so the operation took longer than 5 seconds and showed differences between `torch.rand` and `torch.ones`.  

However, it was still fast, around **35 GB/s**.  

I used an AI to generate this code for synthetically measuring the performance of `torch.rand` vs. `torch.ones`.  
The result is:
```verbatim
mpirun -n 64 python3 pytensortest.py  --gb_per_process 2
Starting PyTorch Tensor Creation Benchmark
MPI World Size: 64 processes
Target data size per process: 2.00 GB
--------------------------------------------------------------------------------------------------------------
Datatype   | Random  | Op Type       | Elements/Rank   | Agg. BW (GB/s)     | Max Time (s) | Data/Rank (GB)
--------------------------------------------------------------------------------------------------------------
fp32       | True    | torch.rand    | 536870912       | 42.44              | 3.016        | 2.00
fp32       | False   | torch.ones    | 536870912       | 108.12             | 1.184        | 2.00
fp16       | True    | torch.rand    | 1073741824      | 18.98              | 6.742        | 2.00
fp16       | False   | torch.ones    | 1073741824      | 118.11             | 1.084        | 2.00
bf16       | True    | torch.rand    | 1073741824      | 15.43              | 8.295        | 2.00
bf16       | False   | torch.ones    | 1073741824      | 117.95             | 1.085        | 2.00
fp64       | True    | torch.rand    | 268435456       | 53.73              | 2.382        | 2.00
fp64       | False   | torch.ones    | 268435456       | 116.20             | 1.102        | 2.00
int8       | True    | torch.randint | 2147483648      | 6.72               | 19.036       | 2.00
int8       | False   | torch.ones    | 2147483648      | 118.01             | 1.085        | 2.00
uint8      | True    | torch.randint | 2147483648      | 6.73               | 19.029       | 2.00
uint8      | False   | torch.ones    | 2147483648      | 118.04             | 1.084        | 2.00
--------------------------------------------------------------------------------------------------------------
Benchmark finished.
```

@zhenghh04  could you try this code using the above mpirun command to simulate the memory allocation for llama70b_zero3? The FP32 line (matching optimizer state of llama70b) is probably the most interesting one. 

```python
import torch
import time
from mpi4py import MPI
import argparse

# Helper function to convert string to torch.dtype (as provided by user)
def get_torch_datatype(datatype_str):
    """Converts a string representation to a torch.dtype object."""
    if datatype_str == "fp32":
        return torch.float32
    elif datatype_str == "fp16":
        return torch.float16
    elif datatype_str == "fp64":
        return torch.float64
    elif datatype_str == "int8":
        return torch.int8
    elif datatype_str == "uint8":
        return torch.uint8
    elif datatype_str == "bf16":  # bfloat16
        return torch.bfloat16
    else:
        raise ValueError(f"Invalid datatype string: {datatype_str}")

# The core tensor creation logic (adapted from user's class method)
def get_tensor_core_op(length, datatype_str_for_lookup, randomize=True):
    """
    Creates a torch tensor with specified characteristics.

    Args:
        length (int): The number of elements in the tensor.
        datatype_str_for_lookup (str): String representation of the data type (e.g., "fp32").
        randomize (bool): If True, fills tensor with random values; otherwise, fills with ones.

    Returns:
        tuple: (torch.Tensor, str) - The created tensor and a description of the operation.
    
    Raises:
        Exception: If randomization is requested for an unsupported data type.
    """
    torch_dtype = get_torch_datatype(datatype_str_for_lookup)
    op_description = ""

    if randomize:
        if torch_dtype in [torch.float32, torch.float16, torch.float64, torch.bfloat16]:
            tensor = torch.rand(length, dtype=torch_dtype)
            op_description = "torch.rand"
        elif torch_dtype == torch.int8:
            # Generates random integers in the range [-128, 127]
            tensor = torch.randint(low=-128, high=128, size=(length,), dtype=torch_dtype)
            op_description = "torch.randint"
        elif torch_dtype == torch.uint8:
            # Generates random integers in the range [0, 255]
            tensor = torch.randint(low=0, high=256, size=(length,), dtype=torch_dtype)
            op_description = "torch.randint"
        else:
            # This path handles dtypes for which randomize=True is not explicitly defined above
            raise ValueError(f"Datatype {torch_dtype} ({datatype_str_for_lookup}) cannot be randomized with current logic.")
    else:
        tensor = torch.ones(length, dtype=torch_dtype)
        op_description = "torch.ones"
    return tensor, op_description

def main():
    """Main function to run the benchmark."""
    parser = argparse.ArgumentParser(description="Benchmark PyTorch tensor creation with MPI.")
    parser.add_argument("--gb_per_process", type=float, default=1.0,
                        help="Target data size in GB to allocate per MPI process (default: 1.0 GB).")
    args = parser.parse_args()

    target_gb_per_process = args.gb_per_process

    # Initialize MPI
    comm = MPI.COMM_WORLD
    rank = comm.Get_rank()
    world_size = comm.Get_size()

    # Datatypes and randomization options to test
    datatypes_to_test = ["fp32", "fp16", "bf16", "fp64", "int8", "uint8"]
    randomize_options = [True, False]

    if rank == 0:
        print(f"Starting PyTorch Tensor Creation Benchmark", flush=True)
        print(f"MPI World Size: {world_size} processes", flush=True)
        print(f"Target data size per process: {target_gb_per_process:.2f} GB", flush=True)
        print("-" * 110, flush=True)
        header = (f"{'Datatype':<10} | {'Random':<7} | {'Op Type':<13} | "
                  f"{'Elements/Rank':<15} | {'Agg. BW (GB/s)':<18} | "
                  f"{'Max Time (s)':<12} | {'Data/Rank (GB)':<15}")
        print(header, flush=True)
        print("-" * 110, flush=True)

    for dt_str in datatypes_to_test:
        for randomize_flag in randomize_options:
            tensor_result_for_size_check = None # To store tensor from rank 0 for size check
            
            try:
                # Determine torch_dtype and element size
                torch_dtype_actual = get_torch_datatype(dt_str)
                # Use a dummy tensor to get element size accurately and safely
                element_size_bytes = torch.tensor([], dtype=torch_dtype_actual).element_size()

                if element_size_bytes == 0: # Should not happen for valid dtypes
                    if rank == 0:
                        print(f"Could not determine element size for {dt_str}. Skipping.", flush=True)
                    continue # Skip to the next iteration

                # Calculate number of elements needed for the target GB
                bytes_per_process = int(target_gb_per_process * (1024**3))
                elements_per_process = bytes_per_process // element_size_bytes

                if elements_per_process == 0:
                    if rank == 0:
                        print(f"Target size {target_gb_per_process:.2f} GB is too small for 1 element of datatype {dt_str}. Skipping.", flush=True)
                    continue # Skip to the next iteration

                # Synchronize all processes before starting the timing for this specific test case
                comm.Barrier()
                
                op_start_time = time.perf_counter()
                
                created_tensor = None
                op_description_str = "N/A"
                error_during_op = False
                
                try:
                    # Each process creates its own tensor
                    created_tensor, op_description_str = get_tensor_core_op(elements_per_process, dt_str, randomize_flag)
                    if rank == 0: # Store tensor from rank 0 for later size check
                        tensor_result_for_size_check = created_tensor

                except Exception as e:
                    error_during_op = True
                    if rank == 0:
                        err_msg = str(e).replace('\n', ' ') # Flatten error message
                        print(f"{dt_str:<10} | {str(randomize_flag):<7} | {'ERROR'.ljust(13)} | "
                              f"{elements_per_process:<15d} | {'N/A'.ljust(18)} | {'N/A'.ljust(12)} | "
                              f"{(bytes_per_process / (1024**3)):<15.2f} | {type(e).__name__}: {err_msg[:40]}...", flush=True)
                
                # Synchronize all processes after the operation to get the time of the slowest process
                comm.Barrier()
                op_end_time = time.perf_counter()

                if error_during_op:
                    # If an error occurred on any process (implicitly handled by rank 0 printing),
                    # other ranks should also skip reporting for this case.
                    continue 

                duration_on_this_rank = op_end_time - op_start_time

                # Gather all durations at the root process (rank 0)
                all_durations = comm.gather(duration_on_this_rank, root=0)

                if rank == 0:
                    if all_durations:
                        max_duration = max(all_durations) # Wall-clock time is dictated by the slowest process
                        
                        # Use the tensor from rank 0 to get actual element size for calculation
                        # This assumes all ranks successfully created tensors of the same type/size
                        if tensor_result_for_size_check is not None:
                            actual_element_size_bytes = tensor_result_for_size_check.element_size()
                            actual_bytes_per_process = elements_per_process * actual_element_size_bytes
                            total_aggregated_bytes = actual_bytes_per_process * world_size
                        else: # Fallback if tensor_result was not captured (should not happen if no error)
                            actual_bytes_per_process = bytes_per_process # Use target, less accurate
                            total_aggregated_bytes = actual_bytes_per_process * world_size


                        aggregated_bandwidth_gb_s = 0.0
                        if max_duration > 1e-9: # Avoid division by zero or tiny numbers
                            aggregated_bandwidth_gb_s = (total_aggregated_bytes / (1024**3)) / max_duration
                        
                        data_per_rank_gb = actual_bytes_per_process / (1024**3)

                        print(f"{dt_str:<10} | {str(randomize_flag):<7} | {op_description_str.ljust(13)} | "
                              f"{elements_per_process:<15d} | {aggregated_bandwidth_gb_s:<18.2f} | "
                              f"{max_duration:<12.3f} | {data_per_rank_gb:<15.2f}", flush=True)
                    else:
                        # This case should ideally not be reached if world_size > 0 and gather works
                        print(f"Warning: No durations gathered for {dt_str}, Randomize: {randomize_flag}", flush=True)
            
            except Exception as e_outer_loop:
                # Catch any unexpected errors in the setup phase for a test case
                if rank == 0:
                    err_msg_outer = str(e_outer_loop).replace('\n', ' ')
                    print(f"Outer loop error for {dt_str}, Randomize: {randomize_flag}: {type(e_outer_loop).__name__}: {err_msg_outer[:60]}... Skipping test case.", flush=True)

    if rank == 0:
        print("-" * 110, flush=True)
        print("Benchmark finished.", flush=True)

if __name__ == "__main__":
    # This ensures that the main function is called only when the script is executed
    # and not when imported as a module.
    main()
```

zhenghh04:
@LouisDDN this is the output 
```
WORLD_SIZE: 24
PPN: 12
PBS_JOBSIZE: 2
Starting PyTorch Tensor Creation Benchmark
MPI World Size: 24 processes
Target data size per process: 2.00 GB
--------------------------------------------------------------------------------------------------------------
Datatype   | Random  | Op Type       | Elements/Rank   | Agg. BW (GB/s)     | Max Time (s) | Data/Rank (GB) 
--------------------------------------------------------------------------------------------------------------
fp32       | True    | torch.rand    | 536870912       | 21.50              | 2.232        | 2.00           
fp32       | False   | torch.ones    | 536870912       | 293.51             | 0.164        | 2.00           
fp16       | True    | torch.rand    | 1073741824      | 9.72               | 4.940        | 2.00           
fp16       | False   | torch.ones    | 1073741824      | 334.97             | 0.143        | 2.00           
bf16       | True    | torch.rand    | 1073741824      | 8.72               | 5.502        | 2.00           
bf16       | False   | torch.ones    | 1073741824      | 331.14             | 0.145        | 2.00           
fp64       | True    | torch.rand    | 268435456       | 25.70              | 1.868        | 2.00           
fp64       | False   | torch.ones    | 268435456       | 331.79             | 0.145        | 2.00           
int8       | True    | torch.randint | 2147483648      | 5.77               | 8.313        | 2.00           
int8       | False   | torch.ones    | 2147483648      | 331.82             | 0.145        | 2.00           
uint8      | True    | torch.randint | 2147483648      | 5.79               | 8.294        | 2.00           
uint8      | False   | torch.ones    | 2147483648      | 341.85             | 0.140        | 2.00           
--------------------------------------------------------------------------------------------------------------
Benchmark finished.
```

LouisDDN:
Thank you, @zhenghh04.  
The minimum rate you obtained is 5GB/s. For 2TB of RAM, which is approximately the maximum per node, it should take ~7 minutes to allocate. It looks acceptable even in the worst-case scenario.  

You mentioned last week that you observed extremely long initialization time. Do you have more details about the test you ran so we can compare with the synthetic numbers you obtained?  
- What test was run?  
- How long did it take to initialize?

PR #300, opened 2025-07-17
Feat/compute model

Druva-D:
Add computation to the models

Todo: Add the other models
- Add the ddp libraries


hariharan-devarajan:
@Druva-D Can we have this PR finalized so that we can merge this. I think this is important behavior to have.

Issue #305, opened 2025-09-17
Tensorflow Symbol Mismatch during Resnet50 and Cosmoflow data generation


Grace-Chang2:
During the data generation process for models resnet50 and cosmoflow, a tensorflow mismatch symbol error continuously seen. The tensorflow version and tensorflow.io version during the pip install of the DLIO tool is tf=2.20.0 and tfio=0.37.1
On initial research it seems the current tensorflow version does not support the tfio=0.37.1 version. So an initial band aid fix that removed this error was to downgrade tensorflow to 2.16.2

(env2) mpl060d:/home/grace/local # **mpirun -np 8 dlio_benchmark workload=resnet50_h100 ++workload.workflow.generate_data=True ++workload.workflow.train=False**

[OUTPUT] 2025-09-17T14:41:04.721413 Running DLIO [Generating data] with 8 process(es)
[OUTPUT] 2025-09-17T14:41:04.724308 Starting data generation
/home/grace/local/env2/lib/python3.12/site-packages/tensorflow_io/python/ops/__init__.py:98: UserWarning: unable to load libtensorflow_io_plugins.so: unable to open file: libtensorflow_io_plugins.so, from paths: ['/home/grace/local/env2/lib/python3.12/site-packages/tensorflow_io/python/ops/libtensorflow_io_plugins.so']
caused by: ['/home/grace/local/env2/lib/python3.12/site-packages/tensorflow_io/python/ops/libtensorflow_io_plugins.so: undefined symbol: _ZN3tsl5mutex6unlockEv']
  warnings.warn(f"unable to load libtensorflow_io_plugins.so: {e}")
/home/grace/local/env2/lib/python3.12/site-packages/tensorflow_io/python/ops/__init__.py:104: UserWarning: file system plugins are not loaded: unable to open file: libtensorflow_io.so, from paths: ['/home/grace/local/env2/lib/python3.12/site-packages/tensorflow_io/python/ops/libtensorflow_io.so']
caused by: ['/home/grace/local/env2/lib/python3.12/site-packages/tensorflow_io/python/ops/libtensorflow_io.so: undefined symbol: _ZNK10tensorflow4data11DatasetBase8FinalizeEPNS_15OpKernelContextESt8functionIFN4absl12lts_202308028StatusOrIN3tsl4core11RefCountPtrIS1_EEEEvEE']
  warnings.warn(f"file system plugins are not loaded: {e}")
[...]

PR #309, opened 2025-10-23
packaging: stringent tensorflow_io dependencies

fortaa:
At this point tensorflow_io is abandonware, and it is not forward-compatible with tensorflow releases. Until it is completely removed the following restrictions apply:

TensorFlow I/O Version, TensorFlow Compatibility
0.37.1, 2.16.x
0.37.0, 2.16.x
0.36.0, 2.15.x
0.35.0, 2.14.x
0.34.0, 2.13.x
0.33.0, 2.13.x

Issue: #305

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,125 +77,19 @@ jobs:
           python -m venv ${VENV_PATH}
           source ${VENV_PATH}/bin/activate
           pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install -r requirements-test.txt
+      - name: test_ai_logging
+        env:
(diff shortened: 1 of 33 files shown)


zhenghh04 (approved):

rayandrew:
@fortaa @zhenghh04 and @hariharan-devarajan 

correct me if I am wrong.
as far as I know we do not use `tensorflow_io` anywhere.
maybe the better solution is to delete this dependency on the requirements.txt and setup.py?

hariharan-devarajan:
@rayandrew I am fine with a newer version. Let's upgrade that and open a new PR. Also can u add a test that can go through that path and actually fail :)

Issue #306, opened 2025-09-17
Slowdown in Data generation with Resnet50 and Cosmoflow models.


Grace-Chang2:
Resnet and Cosmoflow Datagen Performance Regression (tf, tf.io related?) occuring in different servers and systems across operating systems and in different networks.

We're experiencing a severe slowdown in the datagen phase of the benchmarking tool. While a recent TensorFlow symbol mismatch error can be bypassed by downgrading or reinstalling TensorFlow. The data generation speed is now <10% of previous performance. 

Background:

A few weeks ago, running datagen (resnet/cosmoflow) began producing a TensorFlow symbol mismatch error
This could be "resolved" by:
- Downgrading TensorFlow (from 2.20 to 2.16.2) or
- Uninstalling and reinstalling TensorFlow (2.20)
However, once the error is bypassed, the tool executes but at a much slower rate than previously observed.
This error and slowdown behavior is seen across different storage server products and different operating systems.

Environment (A) Info:
OS: Ubuntu 24.04
Python: 3.12.11
TensorFlow: 2.20
Tensorflow.io: 0.37.1
Benchmarking tool version: v2.0

Environment (B) Info:
OS: openSUSE Leap 15.6
Python: 3.12.11
TensorFlow: 2.20
Tensorflow.io: 0.37.1
Benchmarking tool version: v2.0
RAM: 128GB

When running the DLIO mpi datagen command with 4 processes in env B, the write speed is 30MB/s which is incredibly slow for a 128GB RAM server.

hariharan-devarajan:
Typically, this kind of issue is an environment installation problem of conflicting packages.

1. Can you share the installation steps for benchmark you followed. Was it install on a virtual env or .local or root?

2. Can this issue be reproduced in a containerized env? Basically similar setup to what you on a container running the benchmark can help us reproduce and investigate the issue. If you can share a docker script which can reproduce that would help us quickly narrow down the issue.

hariharan-devarajan:
For Datagen slowness, how did you come up with 30MB/s. That is really slow. The resnet dataset is not that large and should be really fast to do. Is it possible that there is slowness in the SSD due to ware and tear? As a quick test can u do a dd command with 1MB block size and 100GB of file and check the bandwidth reported. If that's fast, there is something weird going on while writing tfrecord data. At which point, we can try a different format and see if tfrecord is the problem. If it is the tfrecord, it may be the same env issue as before.

PR #310, opened 2025-10-23
fix: avoid expensive tfrecord2idx forks

fortaa:
After b97baa11 was commited to the DALI repository the data generation throughput is reduced. More so if the script resides on a distributed file system (an order of magnitude of difference).

Issue: #306


hariharan-devarajan (approved):
looks good

zhenghh04 (approved):

PR #313, opened 2025-12-01
remove self.logger from reader factory

karimyoussef91:
Since reader_factory's get_reader was changed to be a static method, the self.logger at line 39 needed to be removed as it throws error "self not defined" when using a custom data reader.


zhenghh04:
@karimyoussef91 Please check this one: https://github.com/argonne-lcf/dlio_benchmark/pull/342.

PR #327, opened 2026-02-27
Pytorch daos dataloader

zhenghh04:


zhenghh04:
@0xE0F Could you fix the conflicts?

hariharan-devarajan (changes requested):
We are missing test cases. I understand we don't have actual doas but we should explore mocking strategies.

enakta:
> We are missing test cases. I understand we don't have actual doas but we should explore mocking strategies.

I think mocking in this case would be mostly testing the mock itself as DaosDataset pretty much hands off samples loading to pytroch.DataLoader.
There should be way to run single node ephemeral DAOS server in the container, let me have a look whether it would work CI.

zhenghh04:
@enakta any update on the mocking?

enakta:
@zhenghh04 waiting for DAOS 2.8 release, then it would be possible to run an ephemeral server for integration tests in a container by just installing a few packages, rather than building a master branch from scratch.

zhenghh04:
> @zhenghh04 waiting for DAOS 2.8 release, then it would be possible to run an ephemeral server for integration tests in a container by just installing a few packages, rather than building a master branch from scratch.

Thanks! We will wait for that release then.

PR #336, opened 2026-03-31
Change all the "GB" and "MB" in printed strings into "GiB" and "MiB" for clarity

FileSystemGuy:
The user didn't have guidance on whether the numbers were base 10 or base 2, so convert to GiB and MiB.


FileSystemGuy:
Created in the wrong repo, this change has value for the master copy of DLIO, but it isn't ready for a PR yet.  Closing this PR.

PR #337, opened 2026-04-03
feat: introduce Multi-Storage Client (MSC) as a storage provider

shunjiad:
## Summary

Adds [NVIDIA Multi-Storage Client (MSC)](https://pypi.org/project/multi-storage-client/) as an **optional** storage provider so DLIO can run data generation, training I/O, and PyTorch checkpointing against MSC-resolved backends (for example object storage such as S3, GCS, Azure, etc), consistent with existing `local_fs`, `parallel_fs`, `s3`, and `aistore` modes.

MSC is integrated for the **indexed binary** dataset format and **PyTorch** checkpointing.

### 1. MSC client configuration

MSC is configured in its own YAML file; see the [Configuration Reference](https://nvidia.github.io/multi-storage-client/references/configuration.html). A minimal **S3** profile looks like this:

```yaml
profiles:
  experiments:
    storage_provider:
      type: s3
      options:
        base_path: my-bucket
        region_name: us-east-1
```

Point MSC at your config file with the **`MSC_CONFIG`** environment variable (for example in the shell before running DLIO):

```bash
export MSC_CONFIG=/path/to/msc_config.yaml
```

### 2. DLIO workload profile (Hydra YAML)

In the workload, set `storage_type` to `msc` and set `storage_root` to an MSC URI. The segment immediately after `msc://` is the **profile name** (a key under `profiles:` in your MSC config, such as `experiments` in the S3 example above); the following path is the prefix for objects under that profile. DLIO passes `storage_root` to `multistorageclient.resolve_storage_client()`, so dataset and checkpoint paths combine with this prefix and are served through MSC.

```yaml
storage:
  storage_type: msc
  storage_root: msc://experiments/megatron-deepspeed/
```

For this snippet, `experiments` must be a profile defined in your MSC configuration file; `megatron-deepspeed/` is the artifact prefix under that profile.

### DIFF
--- a/dlio_benchmark/checkpointing/checkpointing_factory.py
+++ b/dlio_benchmark/checkpointing/checkpointing_factory.py
@@ -42,5 +42,8 @@ class CheckpointingFactory(object):
         elif checkpoint_mechanism_type == CheckpointMechanismType.PT_S3_SAVE:
             from dlio_benchmark.checkpointing.pytorch_s3_checkpointing import PyTorchS3Checkpointing
             return PyTorchS3Checkpointing.get_instance()
+        elif checkpoint_mechanism_type == CheckpointMechanismType.PT_MSC_SAVE:
+            from dlio_benchmark.checkpointing.pytorch_msc_checkpointing import PyTorchMscCheckpointing
+            return PyTorchMscCheckpointing.get_instance()
         else:
             raise Exception(str(ErrorCodes.EC1005))
--- a/dlio_benchmark/checkpointing/pytorch_msc_checkpointing.py
+++ b/dlio_benchmark/checkpointing/pytorch_msc_checkpointing.py
@@ -1,0 +1,79 @@
+"""
+   Copyright (c) 2026, UChicago Argonne, LLC
+   All Rights Reserved
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+"""
+import logging
+import os
+
+try:
+    import multistorageclient as msc
+    MSC_AVAILABLE = True
+except ImportError:
+    MSC_AVAILABLE = False
+    Path = None
+    logging.warning(
+        "Multi-Storage Client (MSC) not available. "
+        "Install with: pip install multi-storage-client"
+    )
+
+from dlio_benchmark.checkpointing.base_checkpointing import BaseCheckpointing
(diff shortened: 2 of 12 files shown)


shunjiad:
@zhenghh04 Could you review the PR? Thanks!

zhenghh04 (commented):
Thank you for this PR — the overall design is clean and consistent with the existing `s3`/`aistore` patterns. The `MockMscClient` with MPI-aware `_merge_from_all_ranks` is elegant, and using `pytest.importorskip` + `extras_require["msc"]` keeps the base install clean. A few issues need to be addressed before merging:

## Critical

### 1. Broken import guard — `MSC_AVAILABLE` in `storage_factory.py` is always `True`

`storage_factory.py` wraps the import of `MscStorage` in a `try/except ImportError`, expecting the import to fail when `multistorageclient` is missing. It does not — `msc_storage.py` catches its own `ImportError` silently and the class definition succeeds regardless. So `MSC_AVAILABLE` is always `True` in `storage_factory.py`, the guard never fires, and the user gets `NameError: name 'msc' is not defined` (from `msc.resolve_storage_client()` inside `__init__`) instead of the friendly message.

Fix: raise in `MscStorage.__init__` directly:

```python
def __init__(self, namespace, framework=None):
    if not MSC_AVAILABLE:
        raise ImportError("multi-storage-client is not installed. pip install multi-storage-client")
```

### 2. Unrelated `setup.py` change removes `tensorflow` from `x86_deps`

`tensorflow>=2.13.1` is silently dropped from `x86_deps` with no explanation. Please split this into a separate PR or add a rationale — it could break existing TF-based users.

## Important

### 3. `walk_node` is non-recursive — `num_subfolders > 0` silently returns empty file list

`client.list(path=p)` returns only immediate children. The regular FS backend uses `os.walk()` which recurses into subdirectories. When users set `num_subfolders_train > 0`, the benchmark silently finds zero files. The test fixture works around this by hardcoding `num_subfolders_train=0`, but the limitation is undocumented. Please either implement recursive listing or add a validation check in `config.py` that rejects `num_subfolders > 0` with `storage_type=msc`.

### 4. `get_uri` uses `os.path.join` — wrong for URI-scheme paths

`os.path.join("msc://foo/", "/absolute")` drops the prefix entirely. Use `posixpath.join` or simple string concatenation with `id.lstrip("/")`.

### 5. `MscStorage.open()` downloads entire file into memory

`client.open()` returns a `BytesIO` containing the full file. For large indexed-binary files this can exhaust per-rank memory. Consider using ranged `get_data(id, None, offset=offset, length=size)` directly in `get_sample`.

## Minor

- **Dead code:** `Path = None` in the `except ImportError` fallback of `msc_storage.py` and `pytorch_msc_checkpointing.py` — remove.
- **Path stripping:** `prefix_path.replace(self.storage.storage_root, "")` in `indexed_binary_msc_reader.py` replaces all occurrences, not just the leading prefix. Use `removeprefix()` (Python 3.9+) or `prefix_path[len(root):]` guarded by `startswith`.
- **Test FileHandler:** `logging.basicConfig(FileHandler("dlio_benchmark_test.log"))` at module level creates a file as a CI side effect. Move to a fixture or drop the file handler.

zhenghh04 (changes requested):
Please see my review comments above.

PR #339, opened 2026-05-14
fix: prevent barrier deadlock with uneven sample distribution

enakta:
When total samples aren't evenly divisible between ranks, number of training steps will be one off for one of the ranks, causing barrier deadlock, due to early exit from the loop.

For instance, if we have 3 ranks, 145 sample files with batch size 16:
- rank 0,1: 49 samples -> 3 steps per epoch,
- rank 2: 47 samples -> 2 steps per epoch (only two full batches)

Ranks 0 and 1 will be waiting for the barrier, while rank 2 will be starting new epoch.

Aligning number of sample files during the initialization solves the issue and makes it a bit more robust agains individual data loader implementations (e.g. PyTroch Dataloader drop_last = True).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,14 +204,14 @@ jobs:
       - name: test_uneven_samples_deadlock
         run: |
           source ${VENV_PATH}/bin/activate
-          mpirun -np 3 pytest -k "test_uneven_samples_deadlock[npz-pytorch-pytorch-13-4]" -v
-          mpirun -np 3 pytest -k "test_uneven_samples_deadlock[npz-pytorch-pytorch-24-4]" -v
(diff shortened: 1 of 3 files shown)


enakta:
The barrier mentioned: https://github.com/argonne-lcf/dlio_benchmark/blob/main/dlio_benchmark/main.py#L343

enakta:
Tests are failing because only 2 slots are available for CI pipeline:

```
There are not enough slots available in the system to satisfy the 3
slots that were requested by the application:
  pytest
Either request fewer slots for your application, or make more slots
available for use.
```
I think it'd would be hard to reproduce the deadlock with only 2 ranks as it never be more than 1 sample difference.

zhenghh04 (approved):

Issue #340, opened 2026-05-25
bug: sample_shuffle=seed causes duplicate samples in file mapping


yanghonggang:
### Description
When `sample_shuffle=seed` is enabled, `build_sample_map_iter` function 
produces duplicate sample reads within files.
### Root Cause
In `config.py:594`, the code uses `sample_list[sample_index] % num_samples_per_file` 
to compute file-local sample index. When sample_list is shuffled, this produces 
incorrect mappings causing duplicates.
### Reproduction
```
$ python -m pytest dlio_sample_shuffle_test.py 
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.12.7, pytest-7.4.4, pluggy-1.0.0
rootdir: /home/hehe/work/dlio_benchmark
configfile: pytest.ini
plugins: anyio-4.2.0, hydra-core-1.3.2
collected 1 item                                                                                                                                                                                           

dlio_sample_shuffle_test.py F                                                                                                                                                                        [100%]

================================================================================================= FAILURES =================================================================================================
_________________________________________________________________________________________ test_sample_shuffle_bug __________________________________________________________________________________________

    def test_sample_shuffle_bug():
        init()
        logging.info("=" * 80)
        logging.info(" DLIO test for sample_shuffle bug reproduction")
        logging.info("=" * 80)
    
        args = ConfigArguments.get_instance()
        args.num_files_train = 2
        args.num_samples_per_file = 4
        args.sample_shuffle = Shuffle.SEED
        args.seed = 42
        args.seed_change_epoch = True
        args.comm_size = 1
        args.my_rank = 0
        args.read_threads = 1
    
        result, _ = args.build_sample_map_iter(["file_0", "file_1"], 8, epoch_number=0)
    
        samples = {"file_0": [], "file_1": []}
        for _, entries in result.items():
            for _, path, idx in entries:
                samples[path.split("/")[-1]].append(idx)
    
        logging.info(f"Result:")
        logging.info(f"  file_0: {samples['file_0']}")
        logging.info(f"  file_1: {samples['file_1']}")
    
        expected = [0, 1, 2, 3]
>       assert (
            sorted(samples["file_0"]) == expected
        ), f"file_0 should have {expected}, got {sorted(samples['file_0'])}"
E       AssertionError: file_0 should have [0, 1, 2, 3], got [0, 1, 1, 3]
E       assert [0, 1, 1, 3] == [0, 1, 2, 3]
[...]

PR #341, opened 2026-05-25
fix: fix duplicate sample reads when sample_shuffle is enabled (#340)

yanghonggang:
The build_sample_map_iter function used `sample_list[sample_index] % num_samples_per_file` to compute file-local sample index, which caused duplicate reads when sample_shuffle was enabled.
Changed to `sample_index % num_samples_per_file` to correctly distribute samples across files regardless of shuffle order.
Added test case to verify the fix.

### DIFF
--- a/dlio_benchmark/utils/config.py
+++ b/dlio_benchmark/utils/config.py
@@ -591,7 +591,7 @@ class ConfigArguments:
                     abs_path = os.path.abspath(file_list[file_index])
                     process_thread_file_map[thread_index].append((sample,
                                                 abs_path,
-                                                sample_list[sample_index] % self.num_samples_per_file))
+                                                sample_index % self.num_samples_per_file))
                     sample_index += 1
                     file_index = (sample_index // self.num_samples_per_file) % num_files
         return process_thread_file_map, samples_sum
--- a/tests/dlio_sample_shuffle_test.py
+++ b/tests/dlio_sample_shuffle_test.py
@@ -1,0 +1,75 @@
+"""
+Copyright (c) 2025, UChicago Argonne, LLC
+All Rights Reserved
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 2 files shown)


yanghonggang:
@zhenghh04 @hariharan-devarajan

zhenghh04 (approved):
The fix is correct. Approving.

**Root cause analysis** (for future reference):

The bug is a birthday-problem collision in the file-local index computation when `sample_shuffle` is enabled.

`build_sample_map_iter` stores `(sample, abs_path, local_idx)` per thread, where `local_idx` is used to seek within the file. File assignment is sequential: positions `0..n-1` → file 0, `n..2n-1` → file 1, etc. (`file_index = sample_index // num_samples_per_file`).

**Before fix:** `local_idx = sample_list[sample_index] % num_samples_per_file`

`sample_list` is a shuffled permutation of global sample indices. Taking mod `n` of shuffled values gives approximately uniform random draws from `{0,...,n-1}`, so collisions follow the birthday problem:

| samples_per_file (n) | P(≥1 collision per file) | Expected fraction duplicated/missing |
|---|---|---|
| 1 | 0% (safe) | 0% |
| 4 | ~91% | ~22% |
| 8 | ~99.8% | ~30% |
| 1000+ | ~100% | ~36.8% (→ 1 - 1/e) |

For typical ML workloads (hundreds–thousands of samples per file), duplicates were **guaranteed on every shuffled epoch**. The existing `samples_sum` checksum in `reconfigure()` sums global sample IDs (not local indices) so it did not catch this — the bug was completely silent.

**After fix:** `local_idx = sample_index % num_samples_per_file`

`sample_index` is a plain monotone counter, so `local_idx` cycles `0→1→2→...→n-1` cleanly for each file. Every file-local position is read exactly once. Thread assignment (`sample_index // samples_per_thread`) and the shuffled sample IDs stored in tuples are both unchanged — only the seek position is fixed.

**Minor nit:** `tests/dlio_sample_shuffle_test.py` adds a `FileHandler` that writes `dlio_sample_shuffle_test.log` to the working directory, which will litter CI runs. Suggest removing it in a follow-up.

PR #342, opened 2026-06-12
fix: replace self.logger with _args.logger in static get_reader

zhenghh04:
Fixes #313.

`get_reader` is decorated with `@staticmethod` so `self` is not defined. Calling `self.logger.info()` raises `NameError` whenever a custom `reader_class` is configured.

**Fix:** replace `self.logger` with `_args.logger` (already in scope as `ConfigArguments.get_instance()`) rather than dropping the log line entirely, so the message is still emitted on rank 0.

The original PR #313 silently dropped the log line. This version preserves the log message with the correct logger.

### DIFF
--- a/dlio_benchmark/reader/reader_factory.py
+++ b/dlio_benchmark/reader/reader_factory.py
@@ -36,7 +36,7 @@ class ReaderFactory(object):
         _args = ConfigArguments.get_instance()
         if _args.reader_class is not None:
             if DLIOMPI.get_instance().rank() == 0:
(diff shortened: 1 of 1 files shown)


zhenghh04:
@hariharan-devarajan , this is a small PR. please review it when you have time.

karimyoussef91:
@hariharan-devarajan , This PR by @zhenghh04 solves the "self.logger" issue in reader class by replacing it with _args.logger instead of removing the logging line. It should fix the issue without suppressing the log.

hariharan-devarajan (approved):
Looks good.

Issue #27, opened 2022-12-15
Post processing issue , ImportError: cannot import name 'quantiles'


karanveersingh5623:
Hi Team , trying to create report from profilers trace , its failing on importing quantiles from statistics module .

```
[root@k8s-worker86 dlio_benchmark]# ll
total 40
drwxr-xr-x  4 root root    54 Dec 15 12:18 configs
-rw-r--r--  1 root root   372 Dec 13 10:48 Dockerfile
drwxr-xr-x  3 root root    69 Dec 13 10:48 docs
-rwxr-xr-x  1 root root     0 Dec 15 12:20 __init__.py
-rw-r--r--  1 root root 11357 Dec 13 10:48 LICENSE
-rw-r--r--  1 root root  8459 Dec 13 10:48 README.md
-rw-r--r--  1 root root  1028 Dec 13 10:48 requirements.txt
-rw-r--r--  1 root root   613 Dec 13 10:48 setup.cfg
-rw-r--r--  1 root root   215 Dec 13 10:48 setup.py
drwxr-xr-x 11 root root   228 Dec 15 12:23 src
drwxr-xr-x  3 root root   106 Dec 13 10:48 tests
[root@k8s-worker86 dlio_benchmark]#
[root@k8s-worker86 dlio_benchmark]#
[root@k8s-worker86 dlio_benchmark]# python3 -m src.dlio_postprocessor.py --output-folder /mnt/dlio/hydra_log/default/2022-12-15-02-27-09
Traceback (most recent call last):
  File "/usr/lib64/python3.6/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib64/python3.6/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/root/dlio_benchmark/src/dlio_postprocessor.py", line 24, in <module>
    from statistics import mean, median, stdev, quantiles
ImportError: cannot import name 'quantiles'
[root@k8s-worker86 dlio_benchmark]#
[root@k8s-worker86 dlio_benchmark]#
[root@k8s-worker86 dlio_benchmark]#
[root@k8s-worker86 dlio_benchmark]# pip install statistics
Requirement already satisfied: statistics in /usr/local/lib/python3.6/site-packages (1.0.3.5)
Requirement already satisfied: docutils>=0.3 in /usr/local/lib/python3.6/site-packages (from statistics) (0.18.1)
[...]

PR #345, opened 2026-06-25
fix(config): propagate dataset.skip_listing through LoadConfig (#504)

russfellows:
## Summary

mlcommons/storage#504 reported infinite S3 listing on RetinaNet (50M
files, >12 h).  PR #27 added a deterministic skip_listing fast path
to `DLIOBenchmark.__init__` (main.py:242-339) and a
`ConfigArguments.skip_listing: bool = False` field (config.py:233).
mlpstorage 3.0.16 auto-injects `++workload.dataset.skip_listing=True`
for every object-storage run.

Despite all three landing, operator reports showed the listing still
ran in full: "expected effect is not observed," even when explicitly
passing `--params dataset.skip_listing=True`.

## Root cause

`LoadConfig()` in `utils/config.py` copies values out of the loaded
Hydra config tree onto `args` — every other dataset field has an
explicit `if 'X' in config['dataset']: args.X = ...` block — but
**`skip_listing` and `listing_validation_interval` were never added**.
The Hydra override reached `config['dataset']`, `LoadConfig` silently
dropped it, `args.skip_listing` stayed `False`, the fast path in
main.py never entered.  No error, no warning, just a 20-minute
listing the operator thought they had disabled.

## Fix

Two missing handlers added between the existing `record_dims` and
`parquet` blocks in `LoadConfig`.  Coercion mirrors what OmegaConf /
Hydra produces in practice:

- Python `bool` (YAML-native true/false) → passed through `bool()`.
- String form (`true`, `True`, `1`, `yes`, `on`) — coerced via
  case-insensitive comparison against a truthy set.  This is what
  `++workload.dataset.skip_listing=true` produces from the CLI.
- Anything else → False.

`listing_validation_interval` goes through `int()` to handle the same
CLI string-coercion case.

## Test plan

- [x] New `tests/test_skip_listing_config.py` — 11 unit tests cover
      every value shape the override path produces (Python bool,
      lowercase string, `'1'`, `'False'`, garbage string, absent
      field, listing_validation_interval as int/string).  All 11 pass.
- [x] Existing fast-CI: 127 tests still pass (test_fast_ci +
      test_issue_regressions + test_virtual_index_map).
- [x] Verified against new s3dlio v0.9.102 wheel:
      \`mlpstorage whatif training unet3d datagen object --params dataset.num_files_train=2\`
      wrote files to MinIO at the deterministic naming convention
      \`img_{N}_of_{T}.npz\` — the same naming the skip_listing fast
      path now correctly generates per-rank.

Refs mlcommons/storage#504

🤖 Generated with [Claude Code](https://claude.com/claude-code)


russfellows:
Closing — opened against wrong repo by mistake. Intended target was mlcommons/DLIO_local_changes. Reopening there.

PR #346, opened 2026-06-26
fix(preflight): verify directory exists and is writable for direct:// and file:// schemes

russfellows:
## Summary

- For `direct://` (O_DIRECT) and `file://` (buffered) URI schemes, the "bucket" is a filesystem path, not an S3 bucket. `_preflight()` previously ran S3 credential and bucket-reachability checks unconditionally, which would always fail for local-filesystem mode even when the path is valid.
- Replace the unconditional S3 checks with a targeted local-filesystem check: verify the storage root **exists as a directory** and **is readable+writable** by the current process. Any failure raises a clear `ValueError` before MPI workers or upload threads start — the same fail-fast guarantee the S3 path provides for object storage.
- S3 credential / endpoint / bucket-reachability checks are skipped entirely for local schemes (no network, no credentials needed).

This change is required to support the `--o-direct` flag being added to mlp-storage (mlcommons/storage#507).

## Test plan

- [ ] `pytest tests/test_odirect_preflight.py -v` — 7 new unit tests all pass:
  - Happy path: `direct://` and `file://` pass for an existing writable directory
  - No S3 credentials required for local schemes
  - `ValueError` raised when directory is missing
  - `ValueError` raised when directory is not writable
  - S3 credential guard still fires for `s3://` scheme
- [ ] Existing test suite: `pytest tests/ -q --ignore=tests/test_odirect_preflight.py`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


russfellows:
Opened against wrong repository by mistake. Correct PR is against mlcommons/DLIO_local_changes.

